PR #7, opened 2023-01-02
[WIP] Implement Unsupervised Tree

### LINKED ISSUE #2: Add unsupervised random (oblique) forest embedding
https://arxiv.org/pdf/1907.02844.pdf

Need to add split criterion, but otw the trees stay the same I believe.

jshinm:
Still WIP. Partially Fixes #2

- This part of PR will lay the foundation codebase for unsupervised tree which will subsequently be built upon to create the unsupervised forest

## Before submitting

## After submitting


adam2392 (commented):
Thanks for starting this. Few comments:

All the scikit-learn files are unnecessary. We can just cimport from sklearn, so you can delete them from this PR.

jshinm:
I see, that makes sense! I'll import our modularized classes by building locally from sklearn working branch until our PRs are merged. Thank you @adam2392!

jshinm:
@adam2392 I think I'm still putting pieces together. With regards to TwoMeans implementation, I need to further develop each method I just pushed. I'm jotting down some thoughts in my mind.
- use `n_outputs` as user defined cluster number
- change `sum_left` and `sum_right` so that they carry child node X
- implement `Node-Wise Feature Generation` in `node_value` as noted in docstring
- think about how proximity matrix will be implemented and change code structure

skmorf/tree/_unsup_criterion.pxd
```diff
@@ -5,6 +5,7 @@
 
 from sklearn.tree._tree cimport SIZE_t, DTYPE_t, DOUBLE_t
 from sklearn.tree._criterion cimport BaseCriterion
+# from ._criterion cimport BaseCriterion
```
adam2392: Remove fluff

skmorf/tree/_unsup_criterion.pxd
```diff
@@ -17,6 +18,7 @@ from sklearn.tree._criterion cimport BaseCriterion
 #
 # Other changes include the removal of "weighted" samples, which is
 # not needed since criterion are compuated on the data itself.
+
```
adam2392: Don't introduce unnecessary LOC changes to keep diff clean.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -70,12 +69,22 @@ cdef class TwoMeans(Criterion):
...
+        self.end = 0
+
+        self.n_samples = 0
+        self.n_node_samples = 0
+        self.weighted_n_node_samples = 0.0
+        self.weighted_n_left = 0.0
+        self.weighted_n_right = 0.0
+
+        cdef SIZE_t k = 0
+        cdef SIZE_t max_n_classes = 0
```
adam2392: There are no classes in unsupervised learning

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -70,12 +69,22 @@ cdef class TwoMeans(Criterion):
...
+
     cdef int init(
         self,
-        const DOUBLE_t[:, ::1] X,
+        const DTYPE_t[:, :] X,
```
adam2392:
```suggestion
        const DTYPE_t[:, ::1] X,
```

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -100,26 +109,44 @@ cdef class TwoMeans(Criterion):
         samples : array-like, dtype=SIZE_t
             A mask on the samples, showing which ones we want to use
         """
-        # need to implement
-        pass
+        self.X = X
+        self.sample_weight = sample_weight
+        self.weighted_n_samples = weighted_n_samples
+        self.sample_indices = sample_indices
```
adam2392:
```suggestion
        self.sample_indices = sample_indices
        
        return 0
```

adam2392: See return type of the function is `int`.

skmorf/tree/_unsup_criterion.pyx
adam2392: This entire snippet should be in the base class right? None of these internal parameters are specific to the type of unsupervised criterion.

Therefore any subclass of `UnsupervisedCriterion` should not need to by default implement `__cinit__` function unless it's like overriding some default behavior.

adam2392: You are also missing instantiations for `sum_total, sum_left, sum_right`. See `_unsup_criterion.pxd` file that defines the interface for `UnsupervisedCriterion`.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -100,26 +109,44 @@ cdef class TwoMeans(Criterion):
...
-        pass
+        self.pos = self.start
+
+        self.weighted_n_left = 0.0
+        self.weighted_n_right = self.weighted_n_node_samples
+        cdef SIZE_t k
+
+        for k in range(self.n_outputs):
+            memset(&self.sum_left[k, 0], 0, self.n_classes[k] * sizeof(double))
+            memcpy(&self.sum_right[k, 0], &self.sum_total[k, 0], self.n_classes[k] * sizeof(double))
```
adam2392: Again UnsupervisedCriterion has no notion of "classes", so irrelevant here. 

However, you need to reset `sum_left` and `sum_right` right?

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -100,26 +109,44 @@ cdef class TwoMeans(Criterion):
...
-        pass
+        self.pos = self.end
+
+        self.weighted_n_left = self.weighted_n_node_samples
+        self.weighted_n_right = 0.0
+        cdef SIZE_t k
+
+        for k in range(self.n_outputs):
+            memset(&self.sum_right[k, 0], 0, self.n_classes[k] * sizeof(double))
+            memcpy(&self.sum_left[k, 0],  &self.sum_total[k, 0], self.n_classes[k] * sizeof(double))
```
adam2392: No notion of class. Instead set `sum_right` and `sum_left` accordingly.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+        # of computations, i.e. from pos to new_pos or from end to new_po.
+        if (new_pos - pos) <= (end - new_pos):
+            for p in range(pos, new_pos):
+                i = sample_indices[p]
+
+                if sample_weight is not None:
+                    w = sample_weight[i]
+
+                for k in range(self.n_outputs):
+                    self.sum_left[k, <SIZE_t> self.y[i, k]] += w
```
adam2392: No notion of class.

adam2392: But `sum_left` should still be updated.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+            self.reverse_reset()
+
+            for p in range(end - 1, new_pos - 1, -1):
+                i = sample_indices[p]
+
+                if sample_weight is not None:
+                    w = sample_weight[i]
+
+                for k in range(self.n_outputs):
+                    self.sum_left[k, <SIZE_t> self.y[i, k]] -= w
```
adam2392: Same comment: https://github.com/neurodata/scikit-tree/pull/7/files#r1064891519

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+
+                self.weighted_n_left -= w
+
+        # Update right part statistics
+        self.weighted_n_right = self.weighted_n_node_samples - self.weighted_n_left
+        for k in range(self.n_outputs):
+            for c in range(self.n_classes[k]):
+                self.sum_right[k, c] = self.sum_total[k, c] - self.sum_left[k, c]
```
adam2392: Same comment: https://github.com/neurodata/scikit-tree/pull/7/files#r1064891519

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+        mu_right = X[pos:end].sum()
+
+        for k in range(pos):
+            left_s += (X[k] - mu_left)**2
+
+        for c in range(pos,end):
+            right_s += (X[c] - mu_right)**2
+
+        impurity = left_s + right_s
+        impurity /= self.weighted_n_node_samples
```
adam2392: `impurity` in TwoMeans criterion is the variance.

Variance formula would be `weight * (left_impurity + right_impurity) / n_samples`.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+
+        cdef double impurity = 0.0
+        cdef double mu_left #left mean
+        cdef double mu_right #right mean
+        cdef double left_c = 0.0 #left cluster
+        cdef double right_c = 0.0 #right cluster
+        cdef SIZE_t k
+        cdef SIZE_t c
+
+        mu_left = X[:pos].sum()
```
adam2392: `.sum()` is a Python function, so this will not work within `nogil`

adam2392: I would take a look at the private attributes of `UnsupervisedCriterion`. (i.e. sum_total, sum_left, sum_right)... and see what values they "should be holding"

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+
+        for c in range(pos,end):
+            right_s += (X[c] - mu_right)**2
+
+        impurity = left_s + right_s
+        impurity /= self.weighted_n_node_samples
+
+        return impurity
 
     cdef void children_impurity(self, double* impurity_left,
```
adam2392: This function will be easier to implement when you have the `_compute_variance()` private Cython function implemented

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
                                 double* impurity_right) nogil:
```
adam2392:
```suggestion
                                double* impurity_right) nogil:
       """Evaluate the impurity in children nodes.

        i.e. the impurity of the left child (samples[start:pos]) and the
        impurity the right child (samples[pos:end]).

        Parameters
        ----------
        impurity_left : double pointer
            The memory address to save the impurity of the left node
        impurity_right : double pointer
            The memory address to save the impurity of the right node
        """
```

adam2392: This docstring may help you.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
         dest : double pointer
             The memory address which we will save the node value into.
         """
-        # need to implement
-        pass
+        cdef SIZE_t k
+
+        for k in range(self.n_outputs):
+            memcpy(dest, &self.sum_total[k, 0], self.n_classes[k] * sizeof(double))
+            dest += self.max_n_classes
```
adam2392: No notion of class. 

Possible tip: See how `node_value` is used in splitter and tree.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+        # impurity_right[0] = entropy_right / self.n_outputs
+        
+        # use left_sum and right_sum to calculate impurity?
+        # or should this method take two separate Xs making up
+        # left and right child nodes?
         pass
 
     cdef void node_value(self, double* dest) nogil:
-        """Compute the node value of samples[start:end] and save it into dest.
+        r"""Compute the node value of samples[start:end] and save it into dest.
```
adam2392: No, it does not handle the computation. See scikit-learn splitter and tree for how `node_value()` is used.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +160,181 @@ cdef class TwoMeans(Criterion):
...
+        projected space. A is generated by randomly sampling from 
+        {-1,+1} lpd times, then distributing these values uniformly 
+        at random in A. l parameter is used to control the sparsity of A
+        and is set to 1/20.
+
+        Each of the d rows \tilde{X}[i; :], i \in {1,2,...d} is then 
+        inspected for the best split point. The optimal split point and 
+        splitting dimension are chosen according to which point/dimension
+        pair minimizes the splitting criteria described in the following 
+        section
```
adam2392: You can move this into the TwoMeans class docstring

adam2392 (changes requested):
Review on generic class methods: Good start, but I left some comments that require you to develop a better understanding of the header and implementation files (pxd and pyx). You should read Cython documentation to understand better if the comments are confusing.

- All methods `init`, `reset`, `reverse_reset` and `update` imo seem generic. There is no specific to `TwoMeans` or `FastBIC` criterion that requires them to be implemented on a case-by-case basis. Therefore they should all be implemented in the `UnsupervisedCriterion` generic base class.

Review on TwoMeans specific methods: Good start, and it's def in the right direction.
impurity in TwoMeans is defined as the Variance. Variance formula would be `impurity = weight * (left_impurity + right_impurity) / n_samples` and then `left_impurity` and `right_impurity` are defined as: `weight * Variance of left child` and `weight * Variance of right child`. So you need to compute variance in `node_impurity` and `children_impurity`. The only thing changing are the pointers.

You should therefore define a separate function to compute variance.

Focus on getting the Criterion correct and ping me w/ questions. We can move onto the splitter once this is done.

adam2392:
> @adam2392 I think I'm still putting pieces together. With regards to TwoMeans implementation, I need to further develop each method I just pushed. I'm jotting down some thoughts in my mind.
> 
> * use `n_outputs` as user defined cluster number

Good idea because rn there is no way of specifying the number of clusters, but technically because it's a tree, we can just control this by controlling the number of leaf nodes, which can be controlled using `max_depth`. Tho this is not exact, so maybe... we will come back to this. 

For now, this is a hard problem and I would just recognize that `n_outputs` is related to supervised learning.

> * change `sum_left` and `sum_right` so that they carry child node X

See comments for hint towards how to do this. See what `memcpy` method is doing for an array of numbers and you're pretty much doing the simpler thing for a single number.

> * implement `Node-Wise Feature Generation` in `node_value` as noted in docstring

See comments. `node_value` does not do the computation.

> * think about how proximity matrix will be implemented and change code structure

We can handle this later on. The proximity matrix is not generated until we hit the tree level.

skmorf/tree/_unsup_criterion.pxd
jshinm: @adam2392 I don't think I wrote this, but I'll keep that in mind. Should I remove these?

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you for the clarification here!

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you @adam2392! I learned so much from this. One question I have is do you have to define `stride` to declare a variable as contiguous? In `_splitter.pyx`, there was a following class variable called X, then `init` would type it as an `object`. My guess was to keep the types the same everywhere, but judging from this usage of ambiguous typing for X, could we go back and forth from contiguous ([:, ::1]) and not contiguous ([:,:])?

```
cdef class BaseDenseSplitter(Splitter):
    cdef const DTYPE_t[:, :] X

    cdef SIZE_t n_total_samples

    cdef int init(
        self,
        object X,
```

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you! I'll watch out for these

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you for this insight, @adam2392! Should I then create `__cinit__` under `UnsupervisedCriterion`? I was modeling after what was in `_criterion.pyx` where they initialize at two levels below the base. For instance, `__cinit__` is in `ClassificationCriterion`
```
cdef class ClassificationCriterion(Criterion):
    """Abstract criterion for classification."""

    def __cinit__(self, SIZE_t n_outputs,
                  cnp.ndarray[SIZE_t, ndim=1] n_classes):
        """Initialize attributes for this criterion.

```

jshinm: Thank you! I'm still piecing together how to define them b/c those are the counts of classes in supervised setting which we don't have in unsupervised (as they are getting that from `y`). I'll resolve this for now and try again in my next revision :smiley:

skmorf/tree/_unsup_criterion.pyx
jshinm: ah! now that I'm reading your comments, this could/should be 1D

skmorf/tree/_unsup_criterion.pyx
jshinm: I'll implement this referring to your comment above

skmorf/tree/_unsup_criterion.pyx
jshinm: oh! Thank you so much for this! I'll do that

skmorf/tree/_unsup_criterion.pyx
jshinm: Awesome! Thank you :smiley:

skmorf/tree/_unsup_criterion.pyx
jshinm: ah I see, thank you!

skmorf/tree/_unsup_criterion.pyx
adam2392: Yeah I think two means and fastBIC would not require any sep logic for initializing C level data structures. So I think sharing cinit should be fine. 

See: http://docs.cython.org/en/latest/src/userguide/special_methods.html#initialisation-methods about cinit and subclasses

skmorf/tree/_unsup_criterion.pyx
adam2392: Oh good catch. Yeah we can leave it as is for now then. We can refactor if upstream sklearn changes it.

skmorf/tree/_unsup_criterion.pxd
adam2392: Oh I see. Then yeah remove these.

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you for the reference!

jshinm:
@adam2392 I left a few TODO comments for my own reference so I can go back to them and revisit. And also, I need to rewrite `_compute_variance` using `sum_total, left, right`

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -15,9 +14,42 @@ cdef class UnsupervisedCriterion(BaseCriterion):
...
+        # See: http://docs.cython.org/en/latest/src/userguide/special_methods.html#initialisation-methods 
+        # about cinit and subclasses
+
+        # TODO: You are also missing instantiations for 
+        # sum_total, sum_left, sum_right. See _unsup_criterion.pxd file 
+        # that defines the interface for UnsupervisedCriterion.
+
+        # TODO: I would take a look at the private attributes of 
+        # UnsupervisedCriterion. (i.e. sum_total, sum_left, sum_right)... 
+        # and see what values they "should be holding"
```
adam2392:
```suggestion
        # Initialize count metric for current, going left and going right
        self.sum_total = 0.0
        self.sum_left = 0.0
        self.sum_right = 0.0
```

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -45,7 +77,7 @@ cdef class UnsupervisedCriterion(BaseCriterion):
         pass
```
adam2392:
```suggestion
        self.sample_weight = sample_weight
        self.samples = samples
        self.start = start
        self.end = end
        self.n_node_samples = end - start
        self.weighted_n_samples = weighted_n_samples
        self.weighted_n_node_samples = 0.0

        cdef SIZE_t i
        cdef SIZE_t p
        cdef DOUBLE_t w = 1.0
        self.sum_total = 0.0

        for p in range(start, end):
            i = samples[p]

            # w is originally set to be 1.0, meaning that if no sample weights
            # are given, the default weight of each sample is 1.0
            if sample_weight != NULL:
                w = sample_weight[i]
            
            # ith sample; overall compute the weighted average
            self.sum_total += w * X[i]
            self.weighted_n_node_samples += w

        # Reset to pos=start
        self.reset()
        return 0
```

They should share the `init(...)` function. I mentioned that all subclasses most likely share this same interface. If you skim the formulas for fastBIC and TwoMeans, there is no real extra data components (that I'm aware of).

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -100,26 +151,50 @@ cdef class TwoMeans(Criterion):
...
+
+        self.weighted_n_left = 0.0
+        self.weighted_n_right = self.weighted_n_node_samples
+        cdef SIZE_t k
+
+        # TODO: Again UnsupervisedCriterion has no notion of "classes", so irrelevant here.
+        # However, you need to reset sum_left and sum_right right?
+        for k in range(self.n_outputs):
+            memset(&self.sum_left[k], 0, sizeof(double))
+            memcpy(&self.sum_right[k], &self.sum_total[k], sizeof(double))
```
adam2392:
```suggestion
        self.sum_left = 0.0
        self.sum_right = self.sum_total
```

You should be able to fill in the other functions, reverse_reset, with the correct implementations. If not, then please write down questions for what you don't understand.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +208,223 @@ cdef class TwoMeans(Criterion):
...
+
+        cdef double impurity = 0.0
+
+        # TODO: impurity in TwoMeans criterion is the variance.
+        # variance = weight * (left_impurity + right_impurity) / n_samples
+        impurity =  self._compute_variance() / self.weighted_n_node_samples
+
+        return impurity
+
+    cdef void _compute_variance(self) nogil:
```
adam2392: Variance function can be written in general as:

```
cdef double _compute_variance(const DTYPE_t[:] Xf,
            double mean,
            SIZE_t* samples,
            DOUBLE_t* sample_weight,
            SIZE_t start, SIZE_t end) nogil:
    """Compute sample variance from vector of feature values.

    Parameters
    ----------
    """
    cdef SIZE_t p
    cdef SIZE_t i
    cdef DOUBLE_t w = 1.0
    cdef double variance = 0.0
    cdef SIZE_t n_samples = end - start

    # compute sample variance
    for p in range(start, end):
        i = samples[p]

        # w is originally set to be 1.0, meaning that if no sample weights
        # are given, the default weight of each sample is 1.0
        if sample_weight != NULL:
            w = sample_weight[i]
        
        # ith sample, and pth feature
        variance += w * (Xf[i] - mean) * (Xf[i] - mean)
    variance /= n_samples
    return variance
```

Please look at this and see either draft an implementation for `children_impurity` and `node_impurity`, or write down question(s) for what you do not understand and why.

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +208,223 @@ cdef class TwoMeans(Criterion):
...
-        # need to implement
-        pass
+        cdef SIZE_t k
+
+        # TODO: no notion of classes
+        # Possible tip: See how node_value is used in splitter and tree.
+        for k in range(self.n_outputs):
+            memcpy(dest, &self.sum_total[k], sizeof(double))
+            # TODO: need to revisit
+            # dest += self.sum_total[k] / self.weighted_n_node_samples
```
adam2392: You should be able to fill in this implementation. It's a one-liner to set the value of `dest`

skmorf/tree/_unsup_criterion.pyx
```diff
@@ -133,35 +208,223 @@ cdef class TwoMeans(Criterion):
...
+            # w is originally set to be 1.0, meaning that if no sample weights
+            # are given, the default weight of each sample is 1.0.
+            if self.sample_weight is not None:
+                w = self.sample_weight[i]
+
+            for k in range(self.n_outputs):
+                X_i = self.X[i]
+                w_X_i = w * X_i
+                self.sum_total[k] += w_X_i
+                self.sq_sum_total += w_X_i * X_i
```
adam2392: When was `sq_sum_total` defined?

adam2392 (changes requested):
I gave you the correct implementations for:

1. variance: use this now in children_impurity and node_impurity
2. I'm not sure if you read the documentation for `memset` and `memcpy`, but those are "setting" and "copying" values over to a C-array. If `sum_left/right/total` is just a number in unsupervised learning, then in our work you just do `self.sum_left = <whatever it should be> (maybe 0, maybe some other number)`.

You should be able to understand what the code is doing and why those changes (if not, you should ask) and then implement:

1. reverse_reset
2. node_impurity
3. children_impurity

adam2392:
@jshinm  Please merge in changes from https://github.com/neurodata/scikit-tree/pull/8. This latest PR I made consolidated the development pipeline to have a modern build system (i.e. meson), which scipy/numpy/etc are all using now. I also enabled all the CI checks, so by the end of this PR, the CIs should all mostly pass.

- rename everything under the `sktree/` module instead of `skmorf`
- look at the meson build instructions in the README. These are instructions to help build and compile the Cython code for you to check the syntax. Lmk if there are any issues.
- Then address comments I left, fix the code structure to the best that you can, fix syntax by running the build commands
- Make a commit and summarize challenges/questions you have

sktree/tree/_unsup_criterion.pxd
```diff
@@ -41,8 +39,8 @@ cdef class UnsupervisedCriterion:
     # Unsupervised criterion can be used with splitter and tree methods.
     cdef int init(
         self,
-        const DOUBLE_t[:, ::1] X,
-        const DOUBLE_t[:] sample_weight,
+        const DOUBLE_t[:,:] X,
+        DOUBLE_t* sample_weight,
```
adam2392: Why are we changing `sample_weight` from a Cython memoryview to pointer?

sktree/tree/_unsup_criterion.pyx
```diff
@@ -103,26 +171,40 @@ cdef class TwoMeans(UnsupervisedCriterion):
         samples : array-like, dtype=SIZE_t
             A mask on the samples, showing which ones we want to use
         """
-        # need to implement
-        pass
+        self.X = X
+        self.sample_weight = sample_weight
+        self.weighted_n_samples = weighted_n_samples
+        self.sample_indices = sample_indices
```
adam2392, line 131: The base class defines `samples`, not `sample_indices`, so this will fail compilation

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +218,223 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        # of computations, i.e. from pos to new_pos or from end to new_po.
+        if (new_pos - pos) <= (end - new_pos):
+            for p in range(pos, new_pos):
+                i = sample_indices[p]
+
+                if sample_weight is not None:
+                    w = sample_weight[i]
+
+                # TODO: sum_left updated
+                for k in range(self.n_outputs):
```
adam2392: Remove notions of `n_outputs` as we discussed. Unless you have questions, lmk.

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +218,223 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+            self.reverse_reset()
+
+            for p in range(end - 1, new_pos - 1, -1):
+                i = sample_indices[p]
+
+                if sample_weight is not None:
+                    w = sample_weight[i]
+
+                # TODO: sum_left updated
+                for k in range(self.n_outputs):
```
adam2392: https://github.com/neurodata/scikit-tree/pull/7/files#r1071549986

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +218,223 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+                    self.sum_left[k] -= w * self.X[i]
+
+                self.weighted_n_left -= w
+
+        # Update right part statistics
+        self.weighted_n_right = (self.weighted_n_node_samples - 
+                                self.weighted_n_left)
+        # TODO: no notion of classes
+        for k in range(self.n_outputs):
```
adam2392: https://github.com/neurodata/scikit-tree/pull/7/files#r1071549986

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +218,223 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+            # w is originally set to be 1.0, meaning that if no sample weights
+            # are given, the default weight of each sample is 1.0.
+            if self.sample_weight is not None:
+                w = self.sample_weight[i]
+
+            for k in range(self.n_outputs):
+                X_i = self.X[i]
+                w_X_i = w * X_i
+                self.sum_total[k] += w_X_i
+                self.sq_sum_total += w_X_i * X_i
```
adam2392: what is sq_sum_total?

sktree/tree/_unsup_criterion.pyx
```diff
@@ -44,7 +67,34 @@ cdef class UnsupervisedCriterion:
...
+            # are given, the default weight of each sample is 1.0
+            if sample_weight != NULL:
+                w = sample_weight[i]
+            
+            # ith sample; overall compute the weighted average
+            self.sum_total += w * X[i]
+            self.weighted_n_node_samples += w
+
+        # Reset to pos=start
+        self.reset()
```
adam2392: `start` and `end` are not defined yet. Thus this type of logic should be elsewhere in `set_sample_pointers` similar to the current open PR inside sklearn.

adam2392 (changes requested):
Please see comments regarding the criterion class.

The full implementation for the criterion does not seem far off based on the code snippets I've sent, so it's good.  Please compile questions and progress update for Thursday's meeting so we can go first over high level questions, then in-depth function by function. Thanks!

skmorf/tree/_unsup_criterion.pyx
jshinm: I set `dest` to sum_total which doesn't need to be deferenced anymore

jshinm: I just removed all the n_outputs in the code so that this becomes a one line of code copying sum_total to dest

skmorf/tree/_unsup_criterion.pyx
jshinm: removed!

sktree/tree/_unsup_criterion.pxd
jshinm: Thank you! I'm not sure what happened here, but I changed it back to memoryview

sktree/tree/_unsup_criterion.pyx
jshinm: Thank you!

sktree/tree/_unsup_criterion.pyx
jshinm: Just removed!

skmorf/tree/_unsup_criterion.pyx
jshinm: Thank you for this :smiley: 

I'll push the code momentarily!

sktree/tree/_unsup_criterion.pyx
jshinm: Thank you! The whole base class `init` seems to be remnants of my initial copy-paste. I made a placeholder for `init` while child classes are fully inheriting cinit from the base class

sktree/tree/_unsup_criterion.pyx
jshinm: Thank you! As mentioned above, I removed the old init from base class. This is handled by `set_sample_pointers`

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +164,165 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        # sum_left from the direction that require the least amount
+        # of computations, i.e. from pos to new_pos or from end to new_po.
+        if (new_pos - pos) <= (end - new_pos):
+            for p in range(pos, new_pos):
+                i = sample_indices[p]
+
+                if sample_weight is not None:
+                    w = sample_weight[i]
+
+                self.sum_left += w * self.X[i]
```
adam2392: `X` is a 2D array, so I think this wouldn't work? Also, what is this LOC trying to do? `w * X[i, :]`?

adam2392: Moreover, the code in scikit-learn for defining `sum_left, sum_right, sum_total`:

```
       # Count labels for each output
        self.sum_total = np.zeros((n_outputs, max_n_classes), dtype=np.float64)
        self.sum_left = np.zeros((n_outputs, max_n_classes), dtype=np.float64)
        self.sum_right = np.zeros((n_outputs, max_n_classes), dtype=np.float64)
```

So sum_left and sum_right have shape # outputs X # of classes. We don't have classes, or # of outputs. In the analogous code for `update()`, the code is:

```
                for k in range(self.n_outputs):
                    self.sum_left[k, <SIZE_t> self.y[i, k]] += w
```

So `sum_left` is just getting values of `w`. What's the idea for adding `w * X[i]`?

sktree/tree/_unsup_criterion.pyx
```diff
@@ -136,35 +164,165 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
 
         Parameters
         ----------
         dest : double pointer
             The memory address which we will save the node value into.
         """
-        # need to implement
-        pass
+        # memcpy(dest, &self.sum_total, sizeof(double))
+        dest = self.sum_total
```
adam2392:
```suggestion
        # assign value of `sum_total` to the memory pointed to by `dest`
        *dest = self.sum_total
```

I would recommend reading up on C++ pointers to just get an idea for what this is doing

adam2392 (changes requested):
I realized there is an issue regarding how the method computes the Criterion. Currently, we pass a read-only memoryview of X. This is analogous to passing in `y` in Supervised setting.

However, in supervised setting, the criterion is computed from the distribution of all y-values between `start` and `end` sample. So even if `y` is multivariate and has additional columns, one would compute the criterion over each column.

In our setting, we are interested in computing the criterion from a given set of feature values (e.g. columns in X), so what we are missing is how to indicate the feature values we want to compute a criterion over. Currently, we only know how to slice rows of X, but not columns of X. We need an indicator for which column of X to compute e.g. TwoMeans (i.e. Variance) on. Note: If we use oblique trees, `X` is first preprocessed with multiplication by a projection matrix, so the end feature vector we are computing a criterion over is not even computable given a column index; we would need the projection matrix.

So there are two options:

1. an additional function that is part of `UnsupervisedCriterion` base class that allows someone to pass in a read-only memoryview of a "1D feature vector" that we would compute the criterion on i.e. something like `criterion.init_feature_vector(Xf[:, k])`
2. alternatively, an additional function that is part of `UnsupervisedCriterion` base class that allows someone to pass in the relevant column index of X that we would compute the criterion on i.e. something like `criterion.set_feature_index(k)`.

I'm inclined towards option 1, since it allows us to generalize later to oblique trees. IN that setting, the splitter handles computing the relevant feature vector, and the criterion class just has a "1D" feature vector that it then computes criteria over as the `start/end` sample changes.

With option 1, the following changes also need to take place:
1. Refactor `init` to not pass in a read-only memoryview of `X` array.
2. Refactor the base criterion class to not have a 2D `X` array attribute. Instead it should be a 1D feature vector that's values can be set using `criterion.set_feature_vector()`

@jshinm can you draft the function described in option 1? It should be fairly quick and maybe 10 LOC at most change. If you have questions, please let me know; this design might change. Currently, I think w/o this change, you will not be able to implement `node_impurity` and `children_impurity`, but all other functions should not be affected.

adam2392:
I've updated the overall branch with changes that I think should work. I realized that the changes that you would need to implement are quite detailed. Instead, hopefully you can take this time to read over the code and verify that you understand what/why/how the changes were made. Sorry I know this process was quite chaotic, but the goal is for you to get experience iterating on the high level design and Cython syntax. 

You should verify, correct any syntax mistakes you see by running the meson build process, and see if you have any questions.

As a next step, I think it'll be helpful for you to implement the FastBIC criterion. I will try to write the rest of the Cython tree and have you review the code, so you can learn why I made any specific changes wrt sklearn code. For the unsupervised tree, we'll have to essentially copy/paste the sklearn Tree/TreeBuilder code because it is not in a state that supports subclassing to do unsupervised learning.

I would skim the relevant Cython files in https://github.com/jshinm/scikit-tree/pull/1/files

jshinm:
Thank you @adam2392! I think I'm getting the high-end overview of the game plan. I'll probably need to review the code some more and reread your comments to actually get the sense of what and where things need to be changed. But as you mentioned, this whole process has been super educational for me. I'm leaving your comments from merged PR so I can refer back to them here

- UnsupervisedCriterion is now modular and will support TwoMeans and FastBIC probably. TwoMeans only requires a few methods to be overridden.
- UnsupervisedSplitter is implemented
- begin UnsupervisedTree
- [Dependencies] Fix poetry file, so that way you can run poetry install and then you will have the relevant dependencies installed. Then you can run the meson build cycle, so that way you can build the Cython files

PR #11, opened 2023-01-25
[DRAFT, ENH] Start working on oblique unsupervised splitting

adam2392:
Signed-off-by: Adam Li <adam2392@gmail.com>

Changes proposed in this pull request:
- implements oblique splits for unsupervised trees

Needs to update the code to use the pointer casting that is done to subclass `SplitRecord`

## Before submitting

- [ ] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

PR #12, opened 2023-02-01
[DRAFT, ENH] Implement FastBIC criterion to unsupervised 

### LINKED ISSUE #6: Implementation of Two-Means and Fast-BIC criterion and unsupervised decision trees
Once `Criterion` and `Splitter` are properly modularized inside scikit-learn, a first step for us to take advantage of this modularity would be to implement the unsupervised decision tree. This will build a tree using only `X`, rather than `y`. 

I have implemented the skeleton/boilerplate for the Two-Means and the base class of `UnsupervisedCriterion`. Scikit-learn supports splitting on both sparse and dense data. For our purposes, we will focus on supporting only dense data for now.

The full solution should be Cython and Python code to implement the unsupervised decision tree using either `TwoMeans` criteria or `FastBIC` criteria. The tasks that need to be implemented are:

#### In Cython

1. [x] Finish implementation of the functions inside `TwoMeans` class inside `_unsup_criterion.pyx`. The main challenges are probably implementing `node_value`. 
2. [x] Implement `UnsupervisedSplitter` inside a new `_unsup_splitter.pyx` and pxd file for the abstract base class for unsupervised splitters.
3. [x] Implement a `BestDenseSplitter` just copying more or less the code from scikit-learn. Note the functionality in scikit-learn doesn't even rely on `y`.
4. [x] Implement an `UnsupervisedTree` class more or less copying a lot of code from scikit-learn
5. [x] Implement an `UnsupervisedTreeBuilder` class and `UnsupervisedBestFirstTreeBuilder` class also more or less copying a lot of code from scikit-learn

#### In Python
1. [x] Implement a `UnsupervisedBaseDecisionTree` class, similar to `BaseDecisionTree`. If we have the issue/PR merged in by then modularizing the Tree class for unsupervised settings, then we can probably inherit from `BaseDecisionTree`. If not, copy for now.
2. [x] Implement a `UnsupervisedRandomForest` class, similar to the `RandomForestClassifier`.
3. [x] Implement unit test cases based on the paper
4. [ ] Re-implement an example doc like https://sporf.neurodata.io/demos/urerf/iris_unsupervised_classification
5. [ ] Additional experiments to make us believe the implementation is correct...

At the end of this PR, there should be the additional Cython files inside `tree/`: `_unsup_splitter.pyx`, `_unsup_splitter.pxd`, `_unsup_tree.pxd`, `_unsup_tree.pyx` and `_unsup_classes.py`.

Note that this package hasn't setup a solid CI/dev build process, so those may be added along the way to make the dev process simpler.

**Additional context**
See: https://arxiv.org/pdf/1907.02844.pdf

@jshinm and @jovo this I think is a good first project for Jong. The Oblique PR will have to wait for a few modularity improvements (I really don't want to copy/paste large tree files unless necessary).

Why are unsupervised decision trees interesting to me? We can use them to estimate conditional mutual information. Rn it has been shown to estimate MI. https://arxiv.org/abs/2110.13883. We want better CMI estimates because they are central for causal discovery algorithms. Note that CMI estimation is a novel idea and would result in a paper once we implement the above.

### LINKED ISSUE #18: Implement FastBIC criterion
> As a subtask of the initial unsupervised RF implementation effort, here we specifically detail timeframe of this task and discuss project

> Finish implementing FastBIC as detailed in Meghana's 2019 paper

**Estimated completion date**
> Current timeline of the project is 1/30/2023 - 2/3/2023 (+2-3 days for documentation/rebranching/cleaning)

jshinm:
Fixes #18

This PR aims to implement FastBIC criterion as detailed in Meghana's 2019 paper. 

## Before submitting

## After submitting


jshinm:
> You are not following the correct API we've discussed. Also use a different branch and PR as I will use this one to develop oblique splits.
> 
> I also don't know what/where BIC_diff and BIC_same is.

@adam2392 I'm a bit lost with the api discussion. Here, I just added a new class for fastBIC like how twomeans is implemented. I'm starting from here b/c I don't know how `BIC_diff_var` and `BIC_same_var` should be handled as shown in Meghana's 2019 paper peudocode,

adam2392:
> @adam2392 I'm a bit lost with the api discussion. Here, I just added a new class for fastBIC like how twomeans is implemented. I'm starting from here b/c I don't know how `BIC_diff_var` and `BIC_same_var` should be handled as shown in Meghana's 2019 paper peudocode,

I see. Can you review the code for TwoMeans then and see where the impurity is computed, and how it is used in `node_impurity` and `children_impurity` and then we can discuss specific questions you have on Thursday's meeting?

jshinm:
The update I mentioned is in `_unsup_criterion.pyx`. Just FYI, I forked from where you did major folder reorg in `oblique` branch. Thank you, Adam!

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +396,159 @@ cdef class TwoMeans(UnsupervisedCriterion):
         self.n_node_samples = end - start
         self.start = start
         self.end = end
+
+cdef class FastBIC(UnsupervisedCriterion):
+    """Placeholder for FastBIC docs
+    """
+
+    cdef double sum_of_squares(
```
adam2392: Since `sum_of_squares` is already implemented in exactly the same way in TwoMeans, we want to perhaps make `sum_of_squares` a separate function that is not part of either class. Then call `sum_of_squares` inside TwoMeans and FastBIC. Note the arguments probably need to be augmented.

We want to keep in mind "DRY".

DRY: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +396,159 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+
+        # If calling without setting the
+        if self.Xf is None:
+            with gil:
+                raise MemoryError(
+                    'Xf has not been set yet, so one must call init_feature_vec.'
+                )
+
+        # first compute mean
+        mean = self.sum_total / n_node_samples
```
adam2392: Should probably normalize with `weighted_n_node_samples` instead of `n_node_samples`. Can you also fix this inside `TwoMeans.node_impurity`?

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +396,159 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+
+        # then compute the variance of the cluster
+        sig = self.sum_of_squares(
+            self.start,
+            self.end,
+            mean
+        ) / self.weighted_n_node_samples
+
+        impurity = -2*n_node_samples*log(n_node_samles/n_samples) \ 
+                    + n_node_samples*log(2*cnp.pi*sig)
```
adam2392: Few comments regarding this line:

1. Does this compile? 
2. Also I find this a bit hard to read. Perhaps, you can simplify.
3. `n_node_samles/n_samples` isn't this just 1?
4. What equation are you following? This does not seem correct when I look at the paper.

adam2392 (commented):
I think the skeleton is correct. The equations do not seem correct.

I know this is just a draft, but if you write new code, I would double check that it at least compiles. You can catch errors this way. The code rn does not compile so hard to review beyond that.

sktree/tree/unsupervised/_unsup_criterion.pyx
jshinm: I see! I thought the same as I was adapting `sum_of_squares` from TwoMeans. I created an issue (#17) to keep track, so that this gets done in an organized manner

sktree/tree/unsupervised/_unsup_criterion.pyx
jshinm: Thank you! I just fixed both instances

sktree/tree/unsupervised/_unsup_criterion.pyx
jshinm: 1. Definitely not especially with the typo here! I wanted to get your quick feedback on the theory side of things before I moved on
2. I'll work on the readability. Perhaps I could create a variable for the prior
3. Please correct me if I'm wrong, but in the paper, I think they are saying (n/N) where n is the current number of sample and N is the total number of samples. I still need to implement `n_samples` but that's kind of the idea that I was following
4. I'll post it in the comment below!

jshinm:
> I think the skeleton is correct. The equations do not seem correct.
> 
> I know this is just a draft, but if you write new code, I would double check that it at least compiles. You can catch errors this way. The code rn does not compile so hard to review beyond that.

Thank you for the immediate feedback @adam2392! I needed to get your quick feedback on the theory side of things b/c I was quite confused with the LOC I was talking about last night where those LOC seem like they will be implemented in the splitter.

Now that I know that the current implementation is what's needed to be assigned to each node, I'll work on the 1. equation 2. compilation of the code.

jshinm:
@adam2392 the equation that adapted is here 

image

which is what's described in the body of the paper as shown below
image

Did I get this wrong?

sktree/tree/unsupervised/_unsup_criterion.pyx
adam2392: `N` is the number of samples considered. At each node, only `n_node_samples` is considered. `n` in your case would be the number of samples determined by the split point. Since this is node_impurity, there is no split point, so `n=N`.

adam2392:
> image

What is w1? and such? It's helpful when sending math to define everything, otw we spend a few iterations just defining what the questions are.

Re the implementation: I think just taking equation 6 as you see in the paper, you can break down two situations.

- for node_impurity: s = 0, so you only have $N * log(2\pi_2 \sigma_2^2) - N * log(\mu_2)$. $\pi_2 = (N-s)/N = 1$, so this is just plugging in whatever you have for the mean and whatever you have for the variance.
- for children_impurity: $s\neq 0$ and you proceed accordingly. 

Based on that, the implementation should be a few LOC. If anything is not clear, please let me know.

jshinm:
> > image
> 
> What is w1? and such? It's helpful when sending math to define everything, otw we spend a few iterations just defining what the questions are.
> 
> Re the implementation: I think just taking equation 6 as you see in the paper, you can break down two situations.
> 
> * for node_impurity: s = 0, so you only have N∗log(2π2σ22)−N∗log(μ2). π2=(N−s)/N=1, so this is just plugging in whatever you have for the mean and whatever you have for the variance.
> * for children_impurity: s≠0 and you proceed accordingly.
> 
> Based on that, the implementation should be a few LOC. If anything is not clear, please let me know.

Thank you so much @adam2392!! I think I now have much better understanding. I'll correct the code accordingly so it compiles then push it

sktree/tree/unsupervised/_unsup_criterion.pyx
jshinm: This is super helpful, thank you so much

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +397,175 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+            end,
+            mean_right
+        ) / self.weighted_n_right
+
+        left_term = log(2*p_l*sig_left/mean_left)
+        right_term = log(2*p_r*sig_right/mean_right)
+
+        # set values at the address pointer
+        impurity_left[0] = s_l*left_term + s_r*right_term
+        impurity_right[0] = s_r*left_term + s_l*right_term
```
jshinm: @adam2392 is this heading towards the right direction?

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +397,175 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        mean = self.sum_total / self.weighted_n_node_samples
+
+        # then compute the variance of the cluster
+        sig = self.sum_of_squares(
+            self.start,
+            self.end,
+            mean
+        ) / self.weighted_n_node_samples
+
+        impurity = n_node_samples*log(2*sig/mean)
```
jshinm, line 463: This is the impurity calculation for note_impurity

adam2392: Yes this looks correct. Can you add an in-line comment to specify what this is computing? (i.e. equation and short explanation)

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +397,175 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        cdef double sig_left
+        cdef double sig_right
+        cdef double left_term
+        cdef double right_term
+
+        # number of samples of left and right
+        s_l = pos - start
+        s_r = end - pos
+
+        # compute prior
```
adam2392:
```suggestion
        # compute prior (i.e. \hat{\pi_1} and \hat{\pi_2} in the paper)
```

Just to clarify

sktree/tree/unsupervised/_unsup_criterion.pyx
adam2392: Can you add an in-line comment specifying the equation each one is computing just for ease of use?

Yes this is the right direction and looks correct to me.

sktree/tree/unsupervised/_unsup_criterion.pyx
```diff
@@ -398,3 +397,175 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        Parameters
+        ----------
+        start : SIZE_t
+            The start sample pointer.
+        end : SIZE_t
+            The end sample pointer.
+        """
+        self.n_node_samples = end - start
+        self.start = start
+        self.end = end
```
adam2392, line 550: Lets move this to the base class and also delete the corresponding function inside `TwoMeans`

adam2392 (changes requested):
Kay this is looking better now!

Now, you will need to add some unit tests and update the docstring within the UnsupervisedRandomForest and UnsupervisedDecisionTree to describe TwoMeans and FastBIC.

adam2392:
FYI I merged in #11 for the oblique tree. I think all you are changing is the criterion two files, so I would just rebase from main (all the extra changes on this branch are unrelated):

- go to main
- checkout a new branch
- copy your criterion changes over
- force push the new branch to the branch of this PR
- delete your old branch associated w/ this PR
- rename your new branch to whatever is pushing to this PR

That shouldn't take more than a few mins, so lmk if you have trouble w/ that.

Alternatively, just start a new PR and close this one.

adam2392:
@jshinm can you migrate your changes into a new clean PR? This one contains a lot of changes that are not from you. It will be easier to review. Thanks!

jshinm:
@adam2392 Thank you for keeping track of my update on the weekend!! That makes sense especially when we are reorganizing directories. I'll push new PR and mention this one from it :smiley:

PR #21, opened 2023-02-03
[ENH] Add basic MORF implementation and design

adam2392:
Towards #20 

Need to also think about the design of the Cython API to allow future possibilities. Also porting over the oblique tree and related examples.

## Before submitting

- [ ] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#making-a-pull-request)
    section of the `CONTRIBUTING` docs.
- [ ] I've updated or added any relevant docstrings following the syntax described in the
    [Writing docstrings](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#writing-docstrings) section of the `CONTRIBUTING` docs.
- [ ] If this PR fixes a bug, I've added a test that will fail without my fix.
- [ ] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [ ] All GitHub Actions jobs for my pull request have passed.

Issue #19, opened 2023-02-02
Create unit test for FastBIC criterion


jshinm:
**Is your feature request related to a problem? Please describe.**
> Create appropriate unit test for FastBIC implementation.

**Describe the solution you'd like**
> This should run as part of meson test cycle

**Estimated completion date**
> Rough timeline for this task is 2/3/2023 - 2/8/2023. This task actually started on 2/5/2023 and expected time of completion is 1-3 days

jshinm:
@adam2392 I've pushed Iris test in this [[PR](https://github.com/jshinm/scikit-tree/pull/3)], but it seems like it builds and tests fine, but it doesn't pass accuracy assertion. Did I misunderstand something here?

![Screenshot from 2023-02-07 00-33-10](https://user-images.githubusercontent.com/52612209/217157043-0173bfae-5289-464e-8ac8-710bb5e19bcc.png)

adam2392:
The issue is that the axis-aligned trees are not as good as the oblique trees. The demo on the iris dataset by default in SPORF used the oblique trees. This requires my changes. You can set the relevant threshold to 0.6 for axis-aligned test instead of 0.9

jshinm:
@adam2392 I ended up bringing it down to 0.3 b/c fastBIC scored 0.34, but with this, though, test_tree builds and tests

adam2392:
Hmm okay... let's see what the forest test does. 

Could be an issue with implementation. But let's wait and see. 

Can we migrate convos regarding PRs into the PR itself?

jshinm:
The changes have been applied and merged into #22

PR #22, opened 2023-02-05
[ENH] Implement FastBIC criterion to unsupervised

### LINKED ISSUE #18: Implement FastBIC criterion

jshinm:
Fixes #18

- This is continuation from #12, where it aims to implement FastBIC criterion as detailed in Meghana's 2019 paper.

## Before submitting

## After submitting

### DIFF
--- a/sktree/tree/_unsup_criterion.pxd
+++ b/sktree/tree/_unsup_criterion.pxd
@@ -54,3 +54,10 @@ cdef class UnsupervisedCriterion(BaseCriterion):
         self,
         const DTYPE_t[:] Xf,
     ) nogil
+
+    cdef double sum_of_squares(
+        self,
+        SIZE_t start,
+        SIZE_t end,
+        double mean
+    ) nogil
--- a/sktree/tree/_unsup_criterion.pxd
+++ b/sktree/tree/_unsup_criterion.pxd
@@ -55,9 +55,8 @@ cdef class UnsupervisedCriterion(BaseCriterion):
         const DTYPE_t[:] Xf,
     ) nogil
 
-    cdef double sum_of_squares(
+    cdef void set_sample_pointers(
         self,
         SIZE_t start,
-        SIZE_t end,
-        double mean
+        SIZE_t end
     ) nogil
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -1,7 +1,5 @@
 #cython: language_level=3, boundscheck=False, wraparound=False, initializedcheck=False, cdivision=True
 
-import numpy as np
-
 cimport numpy as cnp
 from libc.math cimport log
 
@@ -406,7 +404,7 @@ cdef class TwoMeans(UnsupervisedCriterion):
 cdef class FastBIC(TwoMeans):
     r"""Fast-BIC split criterion
     
-    The Bayesian Information Criterion (BIC) is a popular model selection 
+    The Bayesian Information Criterion (BIC) is a popular model seleciton 
     criteria that is based on the log likelihood of the model given data.
 
     Fast-BIC is a method that combines the speed of the two-means clustering 
@@ -465,7 +463,7 @@ cdef class FastBIC(TwoMeans):
         ) / self.weighted_n_node_samples
 
         # simplified equation of maximum log likelihood function at s=0
-        impurity = n_node_samples * log(2 * sig / mean)
+        impurity = n_node_samples*log(2*sig/mean)
 
         return impurity
 
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -498,10 +498,8 @@ cdef class FastBIC(TwoMeans):
         cdef double sig_left
         cdef double sig_left_pre
         cdef double sig_right
-        cdef double BIC_diff_var_l
-        cdef double BIC_diff_var_r
-        # cdef double BIC_same_var_l
-        # cdef double BIC_same_var_r
+        cdef double BIC_diff_var
+        cdef double BIC_same_var
 
         # number of samples of left and right
         s_l = <double>(pos - start)
@@ -534,18 +532,16 @@ cdef class FastBIC(TwoMeans):
 
         # BIC score computed using left and right variances
         # -2(n_1\log{\hat{w}_1}-\frac{n_1}{2}\log{2\pi\hat{\sigma}_{1}^2} - n_2\log{\hat{w}_2}+\frac{n_2}{2}\log{2\pi\hat{\sigma}_{2}^2})
-        BIC_diff_var_l = -2*(s_l*(log(p_l) - log(2*pi*sig_left)/2) + s_r*(-log(p_r) + log(2*pi*sig_right)/2))
-        BIC_diff_var_r = -2*(s_r*(log(p_l) - log(2*pi*sig_left)/2) + s_l*(-log(p_r) + log(2*pi*sig_right)/2))
+        BIC_diff_var = -2*(s_l*(log(p_l) - log(2*pi*sig_left)/2) + s_r*(-log(p_r) + log(2*pi*sig_right)/2))
 
         # BIC score computed using combined variances
         # -2(n_1\log{\hat{w}_1}-\frac{n_1}{2}\log{2\pi\hat{\sigma}_{comb}^2} - n_2\log{\hat{w}_2}+\frac{n_2}{2}\log{2\pi\hat{\sigma}_{comb}^2})
-        BIC_same_var_l = -2*(s_l*(log(p_l) - log(2*pi*sig_comb)/2) + s_r*(-log(p_r) + log(2*pi*sig_comb)/2))
-        BIC_same_var_r = -2*(s_r*(log(p_l) - log(2*pi*sig_comb)/2) + s_l*(-log(p_r) + log(2*pi*sig_comb)/2))
+        BIC_same_var = -2*(s_l*(log(p_l) - log(2*pi*sig_comb)/2) + s_r*(-log(p_r) + log(2*pi*sig_comb)/2))
 
         # simplified equation of maximum log likelihood function 
         # at corresponding sample size for left and right child
-        impurity_left[0] = min(BIC_diff_var_l, BIC_same_var_l)
-        impurity_right[0] = min(BIC_diff_var_r, BIC_same_var_r)
+        impurity_left[0] = min(BIC_diff_var, BIC_same_var)
+        impurity_right[0] = min(BIC_diff_var, BIC_same_var)
 
         # TESTING BELOW
 
@@ -555,18 +551,18 @@ cdef class FastBIC(TwoMeans):
         # printf("sig_left  %f \n", sig_left_pre)
         printf("sig_left  %f \n", sig_left)
         printf("sig_right  %f \n", sig_right)
-        printf("BIC_diff_left  %f \n", BIC_diff_var_l)
-        printf("BIC_diff_right  %f \n", BIC_diff_var_r)
-        # printf("BIC_same_left  %f \n", BIC_same_var_l)
-        # printf("BIC_same_right  %f \n", BIC_same_var_r)
+        printf("sig_comb  %f \n", sig_comb)
+        printf("BIC_diff  %f \n", BIC_diff_var)
+        printf("BIC_same  %f \n", BIC_same_var)
+        printf("impurity_left  %f \n", impurity_left)
+        printf("impurity_right  %f \n", impurity_right)
+        printf("\n\n")
         # printf("BIC_diff_left-zu  %f \n", BIC_diff_var_l)
         # printf("BIC_diff_first_term  %f \n", s_l*(log(p_l) - log(2*pi*sig_left)/2))
         # printf("BIC_diff_second_term  %f \n", s_r*(-log(p_r) + log(2*pi*sig_right)/2))
         # printf("n_node_samples-f  %f \n", self.n_node_samples)
         # printf("n_node_samples-d  %d \n", self.n_node_samples) #int not %f
         # printf("p_l-d  %d \n", s_l / <int>self.n_node_samples)
-        printf("impurity_left  %f \n", impurity_left)
-        printf("impurity_right  %f \n", impurity_right)
         # printf("impurity_left_raw  %f \n", min(BIC_diff_var_l, BIC_same_var_l))
         # printf("impurity_right_raw  %f \n", min(BIC_diff_var_r, BIC_same_var_r))
         # printf("pos  %zu \n", pos)
@@ -577,4 +573,3 @@ cdef class FastBIC(TwoMeans):
         # printf("sample_left-d  %d \n", s_l*2)
         # printf("prior_left  %f \n", p_l)
         # printf("log_left  %f \n", log(2*pi*sig_left))
-        printf("\n\n")
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -406,7 +406,7 @@ cdef class TwoMeans(UnsupervisedCriterion):
 cdef class FastBIC(TwoMeans):
     r"""Fast-BIC split criterion
     
-    The Bayesian Information Criterion (BIC) is a popular model seleciton 
+    The Bayesian Information Criterion (BIC) is a popular model selection 
     criteria that is based on the log likelihood of the model given data.
 
     Fast-BIC is a method that combines the speed of the two-means clustering 
@@ -465,7 +465,7 @@ cdef class FastBIC(TwoMeans):
         ) / self.weighted_n_node_samples
 
         # simplified equation of maximum log likelihood function at s=0
-        impurity = n_node_samples*log(2*sig/mean)
+        impurity = n_node_samples * log(2 * sig / mean)
 
         return impurity
 
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -464,7 +464,8 @@ cdef class FastBIC(TwoMeans):
         ) / self.weighted_n_node_samples
 
         # simplified equation of maximum log likelihood function at s=0
-        impurity = n_node_samples*log(mean/(2*sig))/2 #negative version
+        impurity = n_node_samples*log(2*sig/mean)/2 #negative version
+        # impurity = n_node_samples*log(mean/(2*sig))/2 #negative version
 
         return impurity
 
@@ -508,7 +509,7 @@ cdef class FastBIC(TwoMeans):
 
         # number of samples of left and right
         s_l = pos - start
-        s_r = end - s_l
+        s_r = end - pos#s_l
 
         # compute prior (i.e. \hat{\pi_1} and \hat{\pi_2} in the paper)
         p_l = s_l / self.n_node_samples
@@ -537,8 +538,8 @@ cdef class FastBIC(TwoMeans):
         # right_term = log(2*p_r*sig_right/mean_right)/2
 
         # negative
-        # left_term = log(mean_left/(2*p_l*sig_left))/2
-        # right_term = log(mean_right/(2*p_r*sig_right))/2
+        left_term = log(mean_left/(2*p_l*sig_left))/2
+        right_term = log(mean_right/(2*p_r*sig_right))/2
 
         # pseudocode
         # left_term = log(p_l) - log(2*pi*sig_left)/2
@@ -552,8 +553,8 @@ cdef class FastBIC(TwoMeans):
         # right_term = log(2*p_r*sig_comb/mean_right)/2
 
         # negative
-        left_term = log(mean_left/(2*p_l*sig_left))/2
-        right_term = log(mean_right/(2*p_r*sig_right))/2
+        left_term = log(mean_left/(2*p_l*sig_comb))/2
+        right_term = log(mean_right/(2*p_r*sig_comb))/2
 
         # pseudocode
         # left_term = log(p_l) - log(2*pi*sig_comb)/2
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -2,7 +2,7 @@
 #cython: boundscheck=False, wraparound=False, initializedcheck=False, cdivision=True
 
 cimport numpy as cnp
-from libc.math cimport log
+from libc.math cimport log, pi
 
 cnp.import_array()
 
@@ -464,7 +464,7 @@ cdef class FastBIC(TwoMeans):
         ) / self.weighted_n_node_samples
 
         # simplified equation of maximum log likelihood function at s=0
-        impurity = n_node_samples*log(mean/2*sig)/2
+        impurity = n_node_samples*log(mean/(2*sig))/2 #negative version
 
         return impurity
 
@@ -498,6 +498,13 @@ cdef class FastBIC(TwoMeans):
         cdef double sig_right
         cdef double left_term
         cdef double right_term
+        cdef double BIC_diff_var_l
+        cdef double BIC_diff_var_r
+        cdef double BIC_same_var_l
+        cdef double BIC_same_var_r
+        cdef double BIC_curr_l
+        cdef double BIC_curr_r
+
 
         # number of samples of left and right
         s_l = pos - start
@@ -523,10 +530,42 @@ cdef class FastBIC(TwoMeans):
             mean_right
         ) / self.weighted_n_right
 
-        left_term = log(mean_left/2*p_l*sig_left)/2
-        right_term = log(mean_right/2*p_r*sig_right)/2
+        sig_comb = sig_left + sig_right
+
+        # positive
+        # left_term = log(2*p_l*sig_left/mean_left)/2
+        # right_term = log(2*p_r*sig_right/mean_right)/2
+
+        # negative
+        # left_term = log(mean_left/(2*p_l*sig_left))/2
+        # right_term = log(mean_right/(2*p_r*sig_right))/2
+
+        # pseudocode
+        # left_term = log(p_l) - log(2*pi*sig_left)/2
+        # right_term = -log(p_r) + log(2*pi*sig_right)/2
+
+        BIC_diff_var_l = -2*(s_l*left_term + s_r*right_term)
+        BIC_diff_var_r = -2*(s_r*left_term + s_l*right_term)
+
+        # positive
+        # left_term = log(2*p_l*sig_comb/mean_left)/2
+        # right_term = log(2*p_r*sig_comb/mean_right)/2
+
+        # negative
+        left_term = log(mean_left/(2*p_l*sig_left))/2
+        right_term = log(mean_right/(2*p_r*sig_right))/2
+
+        # pseudocode
+        # left_term = log(p_l) - log(2*pi*sig_comb)/2
+        # right_term = -log(p_r) + log(2*pi*sig_comb)/2
+
+        BIC_same_var_l = -2*(s_l*left_term + s_r*right_term)
+        BIC_same_var_r = -2*(s_r*left_term + s_l*right_term)
+
+        BIC_curr_l = min(BIC_diff_var_l, BIC_same_var_l)
+        BIC_curr_r = min(BIC_diff_var_r, BIC_same_var_r)
 
         # simplified equation of maximum log likelihood function 
         # at corresponding sample size for left and right child
-        impurity_left[0] = s_l*left_term + s_r*right_term
-        impurity_right[0] = s_r*left_term + s_l*right_term
+        impurity_left[0] = BIC_curr_l#-2*(s_l*left_term + s_r*right_term)
+        impurity_right[0] = BIC_curr_r#-2*(s_r*left_term + s_l*right_term)
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -1,7 +1,5 @@
 #cython: language_level=3, boundscheck=False, wraparound=False, initializedcheck=False, cdivision=True
 
-import numpy as np
-
 cimport numpy as cnp
 from libc.math cimport log
 
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -1,4 +1,5 @@
-#cython: language_level=3, boundscheck=False, wraparound=False, initializedcheck=False, cdivision=True
+#cython: language_level=3
+#cython: boundscheck=False, wraparound=False, initializedcheck=False, cdivision=True
 
 cimport numpy as cnp
 from libc.math cimport log
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -558,8 +558,8 @@ cdef class FastBIC(TwoMeans):
         printf("sig_comb  %f \n", sig_comb)
         printf("BIC_diff  %f \n", BIC_diff_var)
         printf("BIC_same  %f \n", BIC_same_var)
-        printf("impurity_left  %f \n", impurity_left)
-        printf("impurity_right  %f \n", impurity_right)
+        printf("impurity_left  %f \n", impurity_left[0])
+        printf("impurity_right  %f \n", impurity_right[0])
         printf("\n\n")
 
         # printf("BIC_diff_left-zu  %f \n", BIC_diff_var_l)
--- a/sktree/tree/_unsup_criterion.pyx
+++ b/sktree/tree/_unsup_criterion.pyx
@@ -470,7 +470,7 @@ cdef class FastBIC(TwoMeans):
         # \text{N}\log{1}-\frac{\text{N}}{2}\log{2\pi\sigma^2}-\frac{\parallel x_{N}-\mu \parallel^2}{2\sigma^2}
         impurity = n_node_samples*log(1)-n_node_samples/2*log(2*pi*sig)-(ss/(2*sig))
 
-        return impurity
+        return -impurity
 
     cdef void children_impurity(
         self,
@@ -498,6 +498,8 @@ cdef class FastBIC(TwoMeans):
         cdef double p_r
         cdef double mean_left
         cdef double mean_right
+        cdef double ss_left
+        cdef double ss_right
         cdef double sig_left
         cdef double sig_right
         cdef double BIC_diff_var_left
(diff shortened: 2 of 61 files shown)


sktree/tree/_unsup_criterion.pyx
```diff
@@ -398,3 +399,154 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        # simplified equation of maximum log likelihood function 
+        # at corresponding sample size for left and right child
+        impurity_left[0] = s_l*left_term + s_r*right_term
+        impurity_right[0] = s_r*left_term + s_l*right_term
+
+    cdef void set_sample_pointers(
+        self,
+        SIZE_t start,
+        SIZE_t end
+    ) nogil:
```
jshinm: @adam2392 should `set_sample_pointers` be moved to base class?

sktree/tree/_unsup_criterion.pyx
```diff
@@ -398,3 +399,154 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+
+        left_term = log(2*p_l*sig_left/mean_left)
+        right_term = log(2*p_r*sig_right/mean_right)
+
+        # simplified equation of maximum log likelihood function 
+        # at corresponding sample size for left and right child
+        impurity_left[0] = s_l*left_term + s_r*right_term
+        impurity_right[0] = s_r*left_term + s_l*right_term
+
+    cdef void set_sample_pointers(
```
adam2392: Move this to unsupervisedcriterion base class

sktree/tree/_unsup_criterion.pyx
```diff
@@ -398,3 +399,154 @@ cdef class TwoMeans(UnsupervisedCriterion):
         self.n_node_samples = end - start
         self.start = start
         self.end = end
+
+cdef class FastBIC(UnsupervisedCriterion):
```
adam2392: Subclass two means criterion instead to leverage reuse of the sum of squares function.

sktree/tree/_unsup_criterion.pyx
```diff
@@ -263,44 +302,6 @@ cdef class TwoMeans(UnsupervisedCriterion):
     pair minimizes the splitting criteria described in the following
     section
     """
-    cdef double sum_of_squares(
```
adam2392, line 265: Don't move this to base class.

sktree/tree/_unsup_criterion.pxd
```diff
@@ -53,4 +53,11 @@ cdef class UnsupervisedCriterion(BaseCriterion):
     cdef void init_feature_vec(
         self,
         const DTYPE_t[:] Xf,
+    ) nogil
+
+    cdef double sum_of_squares(
```
adam2392: Generic unsupervisedcriterions don't need to necessarily compute sum of squares (I.e. variance), so this shouldn't be in the base class.

adam2392: Generic unsupervisedcriterions don't need to necessarily compute sum of squares (I.e. variance), so this shouldn't be in the base class.

adam2392 (changes requested):
Just move the functions back and now add the unit tests to sktree/tree/tests/test_tree.py to finish this off.

sktree/tree/_unsup_criterion.pyx
jshinm: moved this to TwoMeans and subclassed FastBIC to it

sktree/tree/_unsup_criterion.pyx
jshinm: oh you have to submit after you start your review :laughing:

jshinm:
> Just move the functions back and now add the unit tests to sktree/tree/tests/test_tree.py to finish this off.

I'm working on this in a separate PR locally. Can I start a new PR for this for this issue (#19)? It'd be good for my sanity too!

sktree/tree/_unsup_criterion.pyx
```diff
@@ -302,14 +403,45 @@ cdef class TwoMeans(UnsupervisedCriterion):
             ss += w * (self.Xf[s_idx] - mean) * (self.Xf[s_idx] - mean)
         return ss
 
+cdef class FastBIC(TwoMeans):
+    r"""Fast-BIC split criterion
+    
+    The Bayesian Information Criterion (BIC) is a popular model seleciton
```
adam2392, line 410:
```suggestion
    The Bayesian Information Criterion (BIC) is a popular model selection 
```

sktree/tree/_unsup_criterion.pyx
```diff
@@ -323,14 +455,18 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
-        impurity = self.sum_of_squares(
+        # then compute the variance of the cluster
+        sig = self.sum_of_squares(
             self.start,
             self.end,
             mean
         ) / self.weighted_n_node_samples
+
+        # simplified equation of maximum log likelihood function at s=0
+        impurity = n_node_samples*log(2*sig/mean)
```
adam2392:
```suggestion
        impurity = n_node_samples * log(2 * sig / mean)
```

adam2392:
> I'm working on this in a separate PR locally. Can I start a new PR for this for this issue (#19)? It'd be good for my sanity too!

Sure! We can merge in both PRs simultaneously once you demonstrate the unit tests pass and work as expected w/ `'fastbic'` instead of `'twomeans'` in the DecisionTree/Forest.

Not sure why the CIs are not working, but thanks for verifying that the FastBIC criterion compiles. I ran it locally and confirm there is no compilation error.

jshinm:
Ah I see, I'll do that, thank you!

jshinm:
@adam2392 I may have to revert your last rebase. I just spent a few hours trying to debug oblique stuff (which was fun don't get me wrong :smiley:), but my small brain doesn't seem to parallel process two issues at once. One issue I found was that `oblique splitter` was being imported from sklearn instead of sktree which interfered with the build process. If I can't get the oblique unit test working along with current criterion, I'll open a new PR to work on them. I'll keep you updated!

adam2392:
Did you try uninstalling scikit-learn? 

Or deleting the build directory?

jshinm:
It builds fine now after correcting the line for importing obliquesplitrecord from unsup_oblique_tree, which doesn't seem to be related to scikit-learn. And, I build clean every time. 

For testing, test_tree and forest is probably throwing the same errors for both oblique and unsup decision tree, so it's likely something to troubleshoot to get the unit test working for unsup (this didn't happen before my unit test implementation and/or oblique). For now, I've commented out oblique tests/imports to work with 25 errors instead of 50

adam2392:
I'll take a look tomorrow!

adam2392:
This compiles and passes unit-tests on my machine, so it could be an installation issue, or maybe just this errors out on Linux, but not MacOS(?). Do you have the build errors?

Have you updated to `scikit-learn@git+ssh://git@github.com/neurodata/scikit-learn/git@v1.1-refactoredtrees` yet?

Alternatively, I would checkout `tree-featuresv2` branch on the neurodata scikit-learn and build from source to make sure you have any relevant changes.

If those fail, you can consolidate the ObliqueSplitRecord import messages. I am already fixing that by porting over the entire ObliqueTree in #21 because it'll be easier for us to change code here.

jshinm:
@adam2392 I was investing too much time on troubleshooting, so I ended up reverting. It now seems build and test fine. I'll go ahead and make progress with the unit test
 
Screenshot from 2023-02-07 00-00-13

adam2392:
I think actually you can try adding a negative sign when the BIC is computed for child/node. 

In https://github.com/neurodata/scikit-learn/blob/1154502b4a8fef2981c264735541a5ca74c6e764/sklearn/tree/_splitter.pyx#L715-L716, the "best" split point is chosen and the "impurity improvement" is `node - left_child - right_child`. See where the computation is done by the Criterion object: https://github.com/neurodata/scikit-learn/blob/1154502b4a8fef2981c264735541a5ca74c6e764/sklearn/tree/_criterion.pyx#L155-L159. So intuitively impurity is "high" at the root node and as you create splits, the impurity should be "decreasing".

So in twomeans, the impurity score is variance. So as you traverse the tree, intuitively groups of samples should have smaller and smaller variance. 

In fastBIC score, the impurity score is BIC. As you traverse the tree, groups of samples should have larger and larger BIC. Then you need to probably multiply your computation by a negative 1 in order to make the splitter select the right point. (I think this is right. I went over strategies and questions to ask yourself to verify things make sense)

jshinm:
I'll show the comparison here!

jshinm:
@adam2392 I had to spend some time to resolve merge conflict that came with oblique tree update. I previously reverted that rebase to main here, but it showed up again and two codebases got mixed up and failed to build on my end. But it should now be at the state where I've been building and testing

jshinm:
Continuing with the convo on test score, two-means and fastbic performs comparably except on the simulation test for `test_forest` where two-means seems to perform at > 0.6 while fastbic performs at ~0.0

sktree/tree/tests/test_tree.py
```diff
@@ -43,4 +60,27 @@ def test_unsupervisedtree():
     score = adjusted_rand_score(y, predict_labels)
 
     # a single decision tree does not fit well, but should still have a positive score
-    assert score > 0.05
+    assert score >= 0.0
```
jshinm: I changed the threshold from 0.05 to 0.0 b/c fastbic scores ~0.0

adam2392:
> @adam2392 I had to spend some time to resolve merge conflict that came with oblique tree update. I previously reverted that rebase to main here, but it showed up again and two codebases got mixed up and failed to build on my end. But it should now be at the state where I've been building and testing

When that happens, I recommend starting a new branch from main, and puling up your PR diff and copying over relevant changes into a new PR branch. It is difficult to review PRs that are have conflicts w/ main.

adam2392:
I've helped you merge in changes from main

adam2392:
I also recommend keeping a copy of neurodata/scikit-learn fork on `tree-featuresv2` branch that you can always `pip install -e .` from and keep up-to-date.

adam2392:
There are some build errors to sort out because our sklearn-fork dependency needed to get updated w/ latest changes from https://github.com/scikit-learn/scikit-learn/pull/25585/files.

I fixed those in #32 and will merge those in when the CI for the build and unit-tests work. Then we can resolve the build errors on your PR.

To make reviewing something like a discrepancy in scores easier, you should try to make faster iterations and summarize those changes with explicit tables, figures, or screenshots. E.g. I don't know what the exact scores are for "negative", or "not negative" for two-means and fastbic. Those are 4 numbers you can explicitly report in written text in the future.

I'm sorry I don't have any more time to dedicate to this for today. I'll merge #32 later if it works and then if I have time I can help you debug your thing.

adam2392:
I helped merge in the changes from `main`. I also pulled apart the two tests on `blobs` and `iris` so that you can see the explicit failures. In summary there are 8 total tests failing for `fastbic` for both normal unsup tree/forest and oblique unsup tree/forest.

In both `forest` and `tree`, it seems that there is a performance issue for only `fastbic`. I don't think this is due to fastbic being bad, so I suspect there is an error in how `fastbic` is implemented (e.g. in the forest case, with two-means, the oblique unsup forest hits 0.9 score, but super low with fastbic). I think you can probably go thru the paper and the implementation you have line by line and try to figure out if somethings going wrong. 

If you have any questions you can ping me on slack.

jshinm:
@adam2392 Thank you so much!! It doesn't seem like I can build either from main or this branch, so I tried updating sklearn from our `tree-featurev2`, but it still doesn't build. Should I be installing something?

Screenshot from 2023-02-14 12-40-34

Note - I had to use https instead of ssh to get the installation work for sklearn

jshinm:
@adam2392 I just noticed something interesting, and hoping this gives me some lead on where to go next. So I've been re-working the equation, and I noticed that my rearrangement of the terms didn't change the score a bit. So out of curiosity, I just returned 0 for `node_impurity` and `children_impurity` for fastBIC. Turns out, that doesn't change the score either. Then I did the same thing for the 2-means by returning 0, and the test still passes. I built clean every time.

If this is true, could it be possible that these criterions are not actually being utilized?

adam2392:
> @adam2392 I just noticed something interesting, and hoping this gives me some lead on where to go next. So I've been re-working the equation, and I noticed that my rearrangement of the terms didn't change the score a bit. So out of curiosity, I just returned 0 for `node_impurity` and `children_impurity` for fastBIC. Turns out, that doesn't change the score either. Then I did the same thing for the 2-means by returning 0, and the test still passes. I built clean every time.
> 
> If this is true, could it be possible that these criterions are not actually being utilized?

Not sure. You can use `with gil` blocks in Cython `cdef` to print debug statements for yourself, or debug starting from Python code.

adam2392:
> @adam2392 I just noticed something interesting, and hoping this gives me some lead on where to go next. So I've been re-working the equation, and I noticed that my rearrangement of the terms didn't change the score a bit. So out of curiosity, I just returned 0 for `node_impurity` and `children_impurity` for fastBIC. Turns out, that doesn't change the score either. 
> Then I did the same thing for the 2-means by returning 0, and the test still passes. I built clean every time.

Also this is not true. I just did the same thing on `main` and re-ran `sktree/tree/tests` and the tree scores exactly 0.0. For the sake of discussion, what is helpful with questions, or meetings is a **clear presentation** of what you did, why you did it, what was the outcome, and optionally, what you think the outcome means. E.g. you just claimed returning 0 did something. I"m not sure exactly how you did it and what the results were, but they definitely don't line up w/ what I see.

E.g. my results are the following where as on `main` they pass as expected for TwoMeans.

```
FAILED sktree/tree/tests/test_tree.py::test_unsupervisedtree[UnsupervisedDecisionTree] - assert 0.0 > 0.05
FAILED sktree/tree/tests/test_tree.py::test_unsupervisedtree[UnsupervisedObliqueDecisionTree] - assert 0.0 > 0.05
```

jshinm:
@adam2392 Thank you so much for the feedback!! I'll summarize what I've done here.

For fastBIC:
- Arbitrarily made `note_impurity` return 0 for impurity
- Arbitrarily set left and right children to -100 or 0 or 100 in `children_impurity`
- Does not change the output when tested ```dev.py test```

For TwoMeans:
- Arbitrarily made `note_impurity` return 0 for impurity
- Left `children_impurity` untouched
- Still passes the test

Additional checks I've done:
- For TwoMeans, if I set left and right children in `children_impurity` to 0, you are absolutely right that it does not pass the test
- For FastBIC, I've tried setting left and right children to really small numbers like 1e-6, then the number budged little bit
- But for both criterions, whatever the number I put in for the `node_impurity`, the score doesn't seem to change

FastBIC re-attempts:
- I tried different implementations I could find in the paper (i.e., positive, negative)
- Another one I tried is what I could find in the pseudocode section of the paper as below (i.e., BIC_diff_var)
- None of these seems to work

My thoughts on what to do next
- My thought process now is if the rest of the pseudocode needs to be implemented.
- That is, setting the output as BIC_curr, where it's the min(diff_var, same_var).
- One thing I'm not clear though is why it's called same var when you combine vars 

image

jshinm:
@adam2392 I've just pushed all variants of implementation I could come up with from the paper.

1. positive: equation below as is
2. negative: apply minus term to the equation

image

3. pseudocode: equation below as is

image

then I did what's below to compare diff and same, and tested all cases (pos, neg, pseudo). The negative implementation has 1 less error compared to the other two (i.e., 7 errors versus 8)

image

adam2392:
Thanks for the breakdown @jshinm! this is a lot more helpful for me to understand the problem.

Re the disparity between my test and your test with TwoMeans: If you only change `node_impurity`, then that actually doesn't affect the "split points" that are chosen. 

These LOC in sklearn dictate how the "best split" is chosen in a loop. https://github.com/scikit-learn/scikit-learn/blob/4ff92e03d78db5f6eb9abd87594b491cebc80bf9/sklearn/tree/_splitter.pyx#L363-L380.

Then when we look at what proxy impurity is: https://github.com/scikit-learn/scikit-learn/blob/4ff92e03d78db5f6eb9abd87594b491cebc80bf9/sklearn/tree/_criterion.pyx#L148-L164. It is just a function of the children impurities. The node impurity of the parent (i.e. current node before splitting into children) is passed in by reference (https://github.com/scikit-learn/scikit-learn/blob/4ff92e03d78db5f6eb9abd87594b491cebc80bf9/sklearn/tree/_splitter.pyx#L233-L240). Our code is slightly different from this, but essentially the same thing. `node_impurity` is only called once at the root of the tree (you can look at the TreeBuilder code to see where). Then it uses a combination of "proxy impurity improvement" and "children impurity".

Slightly complicated, but tldr that is why changing node_impurity won't affect much besides the initial root node.

Re FastBIC: If you are saying you changed the children impurities and it doesn't do much, then yes that is concerning and implies something is wrong w/ the values being computed. You can insert print statements to print out all the criterion values or something to investigate. 

Reading the pseudocode, it looks similar to what is implemented by the splitter/criterion files, but you can verify yourself by reading the code.

I would double check that the children impurity values are reasonable. Could just be an error in your implementation of the BIC

adam2392:
https://github.com/neurodata/scikit-tree/blob/a7c3b929c8a6cdfd28c5fd12cf9e5a03515db355/sktree/tree/_unsup_criterion.pyx I would try maybe simplifying the formulas a bit. 

A few issues: Seems like end - s_l is incorrect. This assumes `start = 0`. It's a bit hard to read, so perhaps you can map the equations to pieces of code on a piece of paper/presentation and it will help the parse where the error is.

jshinm:
@adam2392 as discussed, I've simplified the equation in its long form comparing left and right BIC score in `child_impurity`. Could you please take a look when you have chance?

With the pseudocode implementation, I'm not sure how it should be applied for `node_impurity`. I gave it a try, but it'd be tremendously helpful if I can get guidance on this as well.

jshinm:
I'm also hoping to build sphinx docs locally, but I can't seem to build scikit-tree. More specifically when I tried this command `sphinx-build -b html docs/ docs/build/html`, the following error shows up. This is from currently up-to-date `main` branch

image

adam2392:
See dev.py --help. 

You can always peruse the CI files to see how it is done.

jshinm:
@adam2392 omg I've been in and out of that menu so many times.. Thank you!

Also for the equation implementation, it may be possible that I'm not coding cython right. For instance, I may be doing something wrong when I'm multiplying size_t (position) by double. I've been playing around with printf, but I think I can come up with a thing or two that I can try

sktree/tree/_unsup_criterion.pyx
```diff
@@ -358,42 +488,56 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
             mean_right
         ) / self.weighted_n_right
 
-    cdef void set_sample_pointers(
-        self,
-        SIZE_t start,
-        SIZE_t end
-    ) nogil:
-        """Set sample pointers in the criterion.
+        sig_comb = sig_left + sig_right
```
adam2392: This is not correct because 1/n(x) + 1/m(y) \neq  1/(n+m) (x+y)

sktree/tree/_unsup_criterion.pyx
```diff
@@ -322,14 +454,18 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
-        impurity = self.sum_of_squares(
+        # then compute the variance of the cluster
+        sig = self.sum_of_squares(
             self.start,
             self.end,
             mean
         ) / self.weighted_n_node_samples
+
+        # simplified equation of maximum log likelihood function at s=0
+        impurity = -2*(n_node_samples*log(1)-log(2*pi*sig)/2)
```
adam2392: See equation 8 if pseudocode

sktree/tree/_unsup_criterion.pyx
jshinm: Thank you! I've pushed an update for this

sktree/tree/_unsup_criterion.pyx
jshinm: Ah I see, I'll reimplement this equation from the appendix

jshinm:
I've just pushed an update fixing dtype issue. Previously, some values threw nan which got fixed after I explicitly assigned dtypes. I also addressed your comment on `sig_comb`. I'll work on the second comment today.

One thing I also need your guidance on is this weird behavior with value assignment, where impurity_left and impurity_right always gets a value from BIC_same_right. I have the feeling that I might be doing something wrong with printf b/c if I comment out printf for BIC_same left and right, impurity left and right gets their values from BIC_diff_right. I'm adding the output for your reference. Please let me know if I can better explain things.

Same code (commenting out printf for BIC_same left and right)
image

Same code (commenting in printf for BIC_same left and right)
image

sktree/tree/_unsup_criterion.pyx
```diff
@@ -358,42 +489,92 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        cdef double sig_left
+        cdef double sig_left_pre
+        cdef double sig_right
+        cdef double BIC_diff_var_l
+        cdef double BIC_diff_var_r
+        # cdef double BIC_same_var_l
+        # cdef double BIC_same_var_r
+
+        # number of samples of left and right
+        s_l = <double>(pos - start)
```
adam2392: Just define with type SIZE_t instead casting here. Same with the other variables. See the definition of those types inside sklearn.

adam2392:
> I've just pushed an update fixing dtype issue. Previously, some values threw nan which got fixed after I explicitly assigned dtypes. I also addressed your comment on `sig_comb`. I'll work on the second comment today.

You should not need to cast. Just define the types properly and there should be no issues. Casting should be used sparingly since there is overhead.

> One thing I also need your guidance on is this weird behavior with value assignment, where impurity_left and impurity_right always gets a value from BIC_same_right. I have the feeling that I might be doing something wrong with printf b/c if I comment out printf for BIC_same left and right, impurity left and right gets their values from BIC_diff_right. I'm adding the output for your reference. Please let me know if I can better explain things.

I don't think the current implementation is correct.

1. https://github.com/neurodata/scikit-tree/blob/d3be43aceed82e1d4bcb9e7c59e77dbd0ccc6453/sktree/tree/_unsup_splitter.pyx#L330-L344 is the computation done by the splitter in lines 24-26 in Algo 3 pseudocode
2. https://github.com/neurodata/scikit-learn/blob/cc16520f156bcdcf8d5613a80810e92f761e12d4/sklearn/tree/_criterion.pyx#L116-L130 computes the criterion as a function of the split point. I.e. it is Line 21, 22, or 23 for FastBIC, or equivalently Line 16 in Algo 2 pseudocode for twomeans.
3. Lines 21-23 is a function of both children, so there is no BIC_diff_var_left, or BIC_diff_var_right. Line 21 and 22 have 4 terms. 2 terms are part of left child and 2 terms are part of the right child. Follow it through and then compute which one gives the min and for the equation that gives the min value (i.e. line 23), you then assign the left and right child impurity values

I would suggest reading through the scikit-learn code from TreeBuilder -> Splitter -> Criterion to understand what phases of the tree each portion handles. Have you managed to do so yet?

jshinm:
@adam2392 To your point 2 and 3, I've changed the BIC_diff_var and BIC_same_var. I'm not sure if I'm following your logic, but having no left and right notion, what I ultimately ended up did was assigning same value for impurity_left and impurity_right. Is this right? 

I've also updated note_impurity implementation following eq. 8 from the paper as you mentioned above. Lastly, I removed casting dtypes.

These are a couple of the intermediate values for variables involved. I feel like it's getting close, but I also feel like I'm going anywhere with this :disappointed: I'll follow through whole process from builder to splitter to criterion.

EDIT - I found out why my impurity_left and right was printing wrong values. I printed out impurity_left instead of impurity[0]. I've updated the stdout. With this correction, it looks like the values at least show normal range of numbers

image

image

Terminal node values
image

adam2392:
> @adam2392 To your point 2 and 3, I've changed the BIC_diff_var and BIC_same_var. I'm not sure if I'm following your logic, but having no left and right notion, what I ultimately ended up did was assigning same value for impurity_left and impurity_right. Is this right?

https://github.com/neurodata/scikit-tree/pull/22#issuecomment-1438593550 

2 terms in each equation are for left child. 2 terms are for right child.

```Python
# these are just different variations of the BIC derivation in the actual paper e.g. equation 6
# compute the following:
- BIC_diff_var_left = function of two terms in line 21
- BIC_same_var_left = function of two terms in line 22
- BIC_diff_var_right = function of two terms in line 21
- BIC_same_var_right = function of two terms in line 22

# next compute the following: see the code I linked in https://github.com/neurodata/scikit-tree/pull/22#issuecomment-1438593550, specifically `criterion.proxy_impurity_improvement`
# I think this is correct. It is whatever results in HIGHER BIC in the children. So prolly want to double check negative signs and such.
if BIC_diff_var_left - BIC_diff_var_right < BIC_same_var_left - BIC_same_var_right:
     left_child = BIC_diff_var_left
     right_child = BIC_diff_var_right
else:
    left_child = BIC_same_var_left
    right_child = BIC_same_var_left
```

Inside `criterion.proxy_impurity_improvement()` is the following:
```Python
left_child - right_child

# or I think exactly it is (i.e. negative minus another number)
- left_child_criterion - right_child_criterion
```

Inside the splitter, we replicate lines 24-26:

```Python
if criterion.proxy_impurity_improvement() > (I think) current_best_criterion:
     current_best_split_point = pos
     current_best_criterion = criterion.impurity_improvement()

# The following is equivalent to the above if I follow the code and substitute exactly what was previously computed:
if left_child - right_child > current_best_criterion:
     current_best_split_point = pos
     current_best_criterion = criterion.impurity_improvement()
```

Lmk what code in this pseudocode or https://github.com/neurodata/scikit-tree/pull/22#issuecomment-1438593550 does not make sense. E.g. TreeBuilder, Splitter, Criterion.

adam2392:
> These are a couple of the intermediate values for variables involved. I feel like it's getting close, but I also feel like I'm going anywhere with this 😞 I'll follow through whole process from builder to splitter to criterion.

Not sure how you're approaching this, but I recommend (on a piece of paper if it helps) to sketch out the process of TreeBuilder, Splitter, and Criterion not paying attention to stuff like types. 

1. What does tree builder do vs what does it tell the Tree to do vs what does it tell the splitter to do?
2. What does the splitter do (e.g. `node_split()`) vs what does it pass off to the criterion?
3. What does the criterion compute and what does it pass back to the splitter to make use of?

Rather than trying to just tackle the whole thing at once, I would recommend to first understand how the whole thing works.

jshinm:
I've split BIC into separate left and right terms. I now have 1 less errors, but I still have 6 errors where my results are as close to 0 as other varied implementation. I'm in and out today, but I'll look into your 3 points.

image
image
image

jshinm:
Forgot to tag @adam2392

sktree/tree/_unsup_criterion.pyx
```diff
@@ -358,42 +492,99 @@ cdef class TwoMeans(UnsupervisedCriterion):
...
+        BIC_diff_var_right = -2*(s_r*(log(p_r) - log(2*pi*sig_right)/2))
+
+        # BIC score computed using combined variances
+        # -2(n_1\log{\hat{w}_1}-\frac{n_1}{2}\log{2\pi\hat{\sigma}_{comb}^2} - n_2\log{\hat{w}_2}+\frac{n_2}{2}\log{2\pi\hat{\sigma}_{comb}^2})
+        BIC_same_var_left = -2*(s_l*(log(p_l) - log(2*pi*sig_comb)/2))
+        BIC_same_var_right = -2*(s_r*(log(p_r) - log(2*pi*sig_comb)/2))
+
+        # simplified equation of maximum log likelihood function 
+        # at corresponding sample size for left and right child
+        if BIC_diff_var_left - BIC_diff_var_right < BIC_same_var_left - BIC_same_var_right:
```
adam2392: You'll need to double check the negative signs all make sense. 

The implementation of splitter and criterion try to create splits that minimizes some criterion metric (Gini, variance, etc.). 

You'll have to make sure your implementation lines up with that.

sktree/tree/_unsup_criterion.pyx
jshinm: I think I fixed this yesterday?

sktree/tree/_unsup_criterion.pyx
jshinm: I will look into it to understand the context. FYI, I tried 1. negative sign 2. switching greater than to less than 3. comparing them to absolute values

sktree/tree/_unsup_criterion.pyx
jshinm: oh this code snippet is outdated

sktree/tree/_unsup_criterion.pyx
adam2392: Oh I mean the children impurities might need to have a negative sign

jshinm: I tried that as well. 1. Negative both 2. Negative left only 3. Negative right only 4. Reversing signs within the two terms that make up left and right children.

jshinm:
> > These are a couple of the intermediate values for variables involved. I feel like it's getting close, but I also feel like I'm going anywhere with this 😞 I'll follow through whole process from builder to splitter to criterion.
> 
> Not sure how you're approaching this, but I recommend (on a piece of paper if it helps) to sketch out the process of TreeBuilder, Splitter, and Criterion not paying attention to stuff like types.
> 
> 1. What does tree builder do vs what does it tell the Tree to do vs what does it tell the splitter to do?
> 2. What does the splitter do (e.g. `node_split()`) vs what does it pass off to the criterion?
> 3. What does the criterion compute and what does it pass back to the splitter to make use of?
> 
> Rather than trying to just tackle the whole thing at once, I would recommend to first understand how the whole thing works.

This is really helpful, thank you, when I get home I'll sit down and work this out

jshinm:
@adam2392 I think the implementation is working! We still have a couple error messages, but it passes most of the major checks like the simulation test requiring 0.9 rand_score. I'll clean up the code & start working on the documentation later today

jshinm:
For screening & setting threshold values for unit test, here are the values I used to update the test_* scripts

```
For forest on simulation data: UnsupervisedRandomForest - twomeans: 0.45748116254036597
For forest on simulation data: UnsupervisedObliqueRandomForest - twomeans: 1.0
For forest on simulation data: UnsupervisedRandomForest - fastbic: 0.9599960816486464
For forest on simulation data: UnsupervisedObliqueRandomForest - fastbic: 1.0
For forest on iris: UnsupervisedRandomForest - twomeans: 0.2104698011643487
For forest on iris: UnsupervisedObliqueRandomForest - twomeans: 0.2217819387664932
For forest on iris: UnsupervisedRandomForest - fastbic: 0.36460228960459423
For forest on iris: UnsupervisedObliqueRandomForest - fastbic: 0.5637510205230709
For tree on simulation data: UnsupervisedDecisionTree - twomeans: 0.09569377990430622
For tree on simulation data: UnsupervisedObliqueDecisionTree - twomeans: 0.09569377990430622
For tree on simulation data: UnsupervisedDecisionTree - fastbic: 0.09569377990430622
For tree on simulation data: UnsupervisedObliqueDecisionTree - fastbic: 0.09569377990430622
For tree on iris data: UnsupervisedDecisionTree - twomeans: 0.011309814500437974
For tree on iris data: UnsupervisedObliqueDecisionTree - twomeans: 0.2101812317413317
For tree on iris data: UnsupervisedDecisionTree - fastbic: 0.00037516827400525235
For tree on iris data: UnsupervisedObliqueDecisionTree - fastbic: 0.00037516827400525235
```

jshinm:
I tried adding this term `if ss == 0.0: ss = 1e-10` in the sum of square to get rid of `inf` and `nan` values, but when I remove them, only the fastBIC fails to work. So I removed that LOC. Maybe some guidance on this?

sktree/tree/_unsup_criterion.pyx
```diff
@@ -299,16 +401,48 @@ cdef class TwoMeans(UnsupervisedCriterion):
                 w = self.sample_weight[s_idx]
 
             ss += w * (self.Xf[s_idx] - mean) * (self.Xf[s_idx] - mean)
+        # if ss == 0.0: ss = 1e-10
```
jshinm, line 404: Here is the LOC to address `int` and 'nan', but then again it breaks fastBIC, so commented out here

adam2392:
> I tried adding this term `if ss == 0.0: ss = 1e-10` in the sum of square to get rid of `inf` and `nan` values, but when I remove them, only the fastBIC fails to work. So I removed that LOC. Maybe some guidance on this?

It is only when taking the log of a zero term that is not defined. Therefore in all the log terms add the same small value of 1e-8 or something. 

This is style review now: Also the BIC equation seems like it is quite repetitive in terms of code. You can simplify it with a function so that way you're not repeating yourself. See DRY: https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself

sktree/tree/_unsup_criterion.pyx
adam2392: This would just make getting zero variance impossible. Zero variance is well defined. 

Log(0) is what is not well defined.

jshinm:
> > I tried adding this term `if ss == 0.0: ss = 1e-10` in the sum of square to get rid of `inf` and `nan` values, but when I remove them, only the fastBIC fails to work. So I removed that LOC. Maybe some guidance on this?
> 
> It is only when taking the log of a zero term that is not defined. Therefore in all the log terms add the same small value of 1e-8 or something.
> 
> This is style review now: Also the BIC equation seems like it is quite repetitive in terms of code. You can simplify it with a function so that way you're not repeating yourself. See DRY: https://en.m.wikipedia.org/wiki/Don%27t_repeat_yourself

Thank you! Before this approach, I tried adding an `alpha` term (1e-10) inside the log term (e.g., log(2*pi*sig_left + alpha)), and threw the same error. 

I looked at 3 terms of the BIC equations, and the 2nd and 3rd terms throws inf and nan. So I didn't add this `alpha` term outside of the log term (e.g., s_r*(log(p_r) - [ log(2*pi*sig_comb)/2 + alpha ] - [ ss_right/(2*sig_comb)) + alpha ]).

adam2392:
I would also double check the prior terms. They divide by integers so maybe they are getting floored. If so just cast one of them to double beforehand. 

What's the "error"?

jshinm:
The first term never goes to inf or nan as far as I could tell (I checked 2nd and 3rd terms of the BIC equations while checking the values for BIC equations where adding that aforementioned LOC removes inf and nan with zero). 

It gives wrong BIC scores thereby doesn't pass the tests. The values are as weird as if I were to implement the equation wrong

PR #53, opened 2023-03-14
Update to devpy 0.1 API

stefanv:
Please note that `devpy` will soon be released as `spin` due to PyPi availability.


adam2392 (approved):
Cool. Thanks for the heads up @stefanv!

PR #57, opened 2023-03-22
ENH: port honest forests from neurodata/honest-forests

### LINKED ISSUE #56: [ENH] Add honest trees
Want to extend current implementations of SPORF and decision trees in scikit-learn to be honest.

Port code from [neurodata/honest-forests](https://github.com/neurodata/honest-forests) into this repo into scikit-tree. Based on issue #29 except specific for the honesty portion.

**Describe alternatives you've considered**
Code written by @rflperry in the [neurodata/honest-forests](https://github.com/neurodata/honest-forests) repo.

**Additional context**
See issue #29

sampan501:
Fixes #56 

- Add honest trees to scikit-tree

## Before submitting

- [x] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#making-a-pull-request)
- [x] I've updated or added any relevant docstrings following the syntax described in the
- [x] If this PR fixes a bug, I've added a test that will fail without my fix.
- [x] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting

- [x] All GitHub Actions jobs for my pull request have passed.


adam2392:
@sampan501 preferably we don't want to have a replica of the "honest tree" for every single type of tree and instead just implement it in one place. In my mind, there are two options:

jw the pros and cons of designing this as a "meta-estimator" instead, which takes any instantiation of a sklearn supervised decision tree (e.g. decisiontreeclassifier, obliquedecisiontreeclassifier, etc.) instead of as a subclass of decisiontreeclassifier? This can then sit in scikit-tree.

Alternatively, we add the honest capabilities to the `DecisionTreeClassifier` and `DecisionTreeRegressor` classes so any supervised tree subclass can enable honesty. This would be a PR to the fork of scikit-learn:tree-featuresv2. Then add the `honest=default_value` to every single tree that inherits from the supervised tree. This is different from what you currently have here. The "honest" feature would be enabled as just a kwarg argument, rather than a separate class.

sampan501:
> instantiation of a sklearn supervised decision tree

Ah, I see what you mean. I'm open to refactoring like this, I just added the initial code that @rflperry wrote as a first step. Here are the two options as I understand them:

1. Have HonestTreeClassifier subclass the respective decision tree type

    * Pros: Just one line of code vs. 2 for the user in the next case. Would have access to all parameters the in the super class.
    * Cons: Much more complex in implementation. Would need to potentially create a different Honest Forest for each classifier. Also (potentially) difficult for the user to use since they would have to parse the documentation to find which one they need

2. Have HonestTreeClassifier as a "meta-class" that takes in the respective decision tree as a parameter:

    * Pros: Much simpler in implementation and much easier for the user. Only one additional line of code that the user has to write (instantiating the object) and the user just has to pic the tree type.
    * Cons: What is the base class we inherit from? We have lost access to parameters the user inputs when instantiating the object that are not stored as class attributes.

3. Add an honesty flag to scikit-learn:tree-featuresv2

    * Pros: Easiest for the user (just an additional flag). Possibly the fastest method (whether or not it is easiest to maintain is a different matter)
    * Cons: Most difficult implementation (would have to parse the Cython and translate). Each implementation would be dependent on scikit-learn:tree-featuresv2. So, if this fork it not updated regularly, then potential new trees from scikit-learn main cannot be implemented. Also, if scikit-learn:tree-featuresv2 ever ends up getting merged with scikit-learn main, we are dependent on them accepting our PRs for new tree types.

Just some that I thought of. Based on this, seems like option 2 is better? Should I create a new module or store everything in "ensemble"

adam2392:
> > instantiation of a sklearn supervised decision tree
> 
> Ah, I see what you mean. I'm open to refactoring like this, I just added the initial code that @rflperry wrote as a first step. Here are the two options as I understand them:
> 
> 1. Have HonestTreeClassifier subclass the respective decision tree type

This is what you currently have in the PR. The second option I was floating is actually to refactor this PR and add the code to both `DecisionTreeClassifier` and `DecisionTreeRegressor` classes in the fork of sklearn. I edited my comment in https://github.com/neurodata/scikit-tree/pull/57#issuecomment-1480099420. This is the most general approach and imo is "cleaner", but might bit a bit more work since we want to keep the fork very well maintained wrt the sklearn internal design.

> 2. Have HonestTreeClassifier as a "meta-class" that takes in the respective decision tree as a parameter:
> Just some that I thought of. Based on this, seems like option 2 is better? Should I create a new module or store everything in "ensemble"

If we do option 2, then I would create a new class in `tree/_classes.py` and also a new class in `ensemble/_supervised_forest.py`. There are two approaches to implementing this. One is to use the `MetaEstimatorMixin` from sklearn, which means we follow the design:

```
class HonestTree(MetaEstimatorMixin):
     def __init__(estimator, honest_ratio=0.5):
         ...

# example
clf = HonestTree(DecisionTreeClassifier())
```
where the estimator is pre-defined. When we do fit, we probably just check that `estimator` is of type `BaseDecisionTree`. The cons with this approach is that this is a completely different API then before.

Alternatively, we would use the design:

```
class HonestTree(BaseEstimator):
     def __init__(tree=DecisionTreeClassifier, honest_ratio=0.5, other tree kwargs...):
         ...

# example
clf = HonestTree(DecisionTreeClassifier, max_feature=...)
```
where we pass in kwargs. The cons with this approach is that the user has to make sure they pass in all the parameters they want. E.g. a ObliquDecisiionTreeClassifier has more parameters than a normal DecisionTreeClassifier.

To ensure stability wrt the underlying class, there would need to be some error-checking of the tree instance.

These aren't necessarily fatal issues, but they do make life a little bit less ideal, so I'm slightly in favor of the option to add the honesty functionality to the fork of sklearn. However, I can defer to whatever one you think is easiest to implement and maintain.

Lmk if this doesn't make sense and we can have a quick chat.

adam2392:
> Add an honesty flag to scikit-learn:tree-featuresv2
Pros: Easiest for the user (just an additional flag). Possibly the fastest method (whether or not it is easiest to maintain is a different matter)
Cons: Most difficult implementation (would have to parse the Cython and translate). Each implementation would be dependent on scikit-learn:tree-featuresv2. So, if this fork it not updated regularly, then potential new trees from scikit-learn main cannot be implemented. Also, if scikit-learn:tree-featuresv2 ever ends up getting merged with scikit-learn main, we are dependent on them accepting our PRs for new tree types.

We can just ignore the Cython part. Just do everything in Python exactly like you're doing now, where we just modify the fit function in `DecisionTreeClassifier` and `DecisionTreeRegressor`.

sampan501:
Based on talk, we decided on option 2

adam2392:
> Based on talk, we decided on option 2

And also on the following design option, with some checks on type of tree (must be classifier or regressor), and if it's already fitted or not:

```
class HonestTree(MetaEstimatorMixin):
     def __init__(estimator, honest_ratio=0.5):
         ...

# example
clf = HonestTree(DecisionTreeClassifier())
```

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,497 @@
...
+    >>> clf = HonestForestClassifier(max_depth=2, random_state=0)
+    >>> clf.fit(X, y)
+    HonestForestClassifier(...)
+    >>> print(clf.predict([[0, 0, 0, 0]]))
+    [1]
+    """
+
+    def __init__(
+        self,
+        n_estimators=100,
```
adam2392, line 314: We may want to allow the passing in of a tree estimator with the default being decisiontreeclassifier?

adam2392 (commented):
The honest tree meta estimator LGTM. Wanna add a unit test, or port the unit test if it already exists?

Also it'll be good to add a sklearn check_estimator test to the classes to check if we missed any edge case behavior.

sktree/ensemble/_honest_forest.py
sampan501: Yep haven't made edits to the forest yet, that commit was just the tree

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,295 @@
+"""Module for forest-based estimators"""
+# Authors: Ronan Perry
+# Adopted from: https://github.com/rflperry/ProgLearn/blob/UF/
+# License: MIT
```
adam2392: I would remove this, since we have a BSD-3 license file in root

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,295 @@
...
+class HonestForestClassifier(ForestClassifier):
+    """
+    A random forest classifier with honest leaf estimates.
+
+    A random forest is a meta estimator that fits a number of decision tree
+    classifiers on various sub-samples of the dataset and uses averaging to
+    improve the predictive accuracy and control over-fitting.
+    The sub-sample size is controlled with the `max_samples` parameter if
+    `bootstrap=True` (default), otherwise the whole dataset is used to build
+    each tree.
```
adam2392: We can remove this.

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,295 @@
...
+from sktree.tree import HonestTreeClassifier
+from sklearn.utils.validation import check_is_fitted
+from joblib import Parallel, delayed
+from sklearn.ensemble._base import _partition_estimators
+import threading
+
+
+class HonestForestClassifier(ForestClassifier):
+    """
+    A random forest classifier with honest leaf estimates.
```
adam2392:
```suggestion
    A forest classifier with honest leaf estimates.
```

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,295 @@
...
+        might contain NaN. This attribute exists when `honest_score_` is True.
+
+    structure_indices_ : list of lists, shape=(n_estimators, n_structure)
+        Indices of training samples used to learn the structure
+
+    honest_indices_ : list of lists, shape=(n_estimators, n_honest)
+        Indices of training samples used to learn leaf estimates
+
+    Notes
+    -----
```
adam2392, line 263: We can mention in the docstring `Notes` something along these lines:

```
Honesty is a feature of trees that enables unbiased estimates of confidence intervals(?) <some description>
with citation to Athey's paper(s). The default implementation here is using double sampling to implement honesty. The amount of samples used for learning split nodes vs leaf nodes is controlled by the ``honest_fraction`` parameter.

This forest classifier is a "meta-estimator" because any tree model can be used in the classification process,
while enabling honesty separates the data used for split and leaf nodes.
```

adam2392: Or some paraphrased version of that.

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def __init__(
+        self,
+        estimator=None,
+        honest_fraction=0.5,
+        honest_prior="empirical",
+    ):
+        if not estimator:
+            self.estimator_ = DecisionTreeClassifier()
+        self.estimator_ = estimator
```
adam2392: You can migrate this to within the `fit` function and then set `self.estimator=estimator`.

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+        nonzero_indices = np.where(sample_weight > 0)[0]
+
+        self.structure_indices_ = np.random.choice(
+            nonzero_indices,
+            int((1 - self.honest_fraction) * len(nonzero_indices)),
+            replace=False,
+        )
+        self.honest_indices_ = np.setdiff1d(nonzero_indices, self.structure_indices_)
+
+        sample_weight[self.honest_indices_] = 0
```
adam2392: This looks like it can go after the call to `estimator_.fit`

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+            Allow to bypass several input checking.
+            Don't use this parameter unless you know what you do.
+
+        Returns
+        -------
+        self : HonestTreeClassifier
+            Fitted estimator.
+        """
+        if check_input:
+            X, y = check_X_y(X, y)
```
adam2392, line 316: By calling `estimator_.fit` first, this check is done internally by scikit-learn's baseDecisionTree, so we can remove it.

adam2392: Instead, I would add a check here for `if self.estimator is not of type BaseDecisionTree: <raise some errors>`

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+        # update the number of classes, unsplit
+        # self.n_samples_, self.n_features_in_ = X.shape
+        if y.ndim == 1:
+            # reshape is necessary to preserve the data contiguity against vs
+            # [:, np.newaxis] that does not.
+            y = np.reshape(y, (-1, 1))
+        check_classification_targets(y)
+        y = np.copy(y).astype(int)
+        # Normally called by super
+        X = self.estimator_._validate_X_predict(X, True)
```
adam2392, line 377: Is this necessary since validation occurs in `estimator_.fit`?

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+            int((1 - self.honest_fraction) * len(nonzero_indices)),
+            replace=False,
+        )
+        self.honest_indices_ = np.setdiff1d(nonzero_indices, self.structure_indices_)
+
+        sample_weight[self.honest_indices_] = 0
+
+        # Learn structure on subsample
+        self.estimator_.fit(
+            X,
```
adam2392, line 356: Should this be `X[self.structure_indices_,:]` and `y[self.structure_indices_, :]`?

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+        check_classification_targets(y)
+        y = np.copy(y).astype(int)
+        # Normally called by super
+        X = self.estimator_._validate_X_predict(X, True)
+        # Fit leaves using other subsample
+        honest_leaves = self.tree_.apply(X[self.honest_indices_])
+
+        self.tree_.value[:, :, :] = 0
+        for leaf_id, yval in zip(honest_leaves, y[self.honest_indices_, 0]):
+            self.tree_.value[leaf_id][0, yval] += 1
```
adam2392: This can be migrated to a private function perhaps called `_set_leaf_nodes(self, X, y)`, where (X, y) are a data pair where X is used to traverse the already built tree and y is used to set the leaf nodes. 

This can then lead us to easily supporting `PropensityTree`

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+    n_features_in_ : int
+        Number of features seen during :term:`fit`.
+
+        .. versionadded:: 0.24
+
+    feature_names_in_ : ndarray of shape (`n_features_in_`,)
+        Names of features seen during :term:`fit`. Defined only when `X`
+        has feature names that are all strings.
+
+        .. versionadded:: 1.0
```
adam2392: Remove.

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,338 @@
...
+
+    n_classes_ : int or list of int
+        The number of classes (for single output problems),
+        or a list containing the number of classes for each
+        output (for multi-output problems).
+
+    n_features_in_ : int
+        Number of features seen during :term:`fit`.
+
+        .. versionadded:: 0.24
```
adam2392: Remove.

adam2392 (commented):
Left a few comments. Lmk what you think.

adam2392:
Thinking of making some large scale changes sometime in the next few weekends: https://github.com/scikit-learn-contrib/scikit-learn-contrib/issues/61#issuecomment-1512881254

Just checking in on this to see if you're able to get this in to consolidate the repo before then. Ideally aiming to merge this and #64 and #65 before then to make life easier

adam2392:
@sampan501 just checking in. what else is there to do on this PR?

sampan501:
I have to fix the ci errors. I plan on finishing that after submitting the supplemental to neurips

sampan501:
Quick question: you're not planning on using hyppo as a dependency of scikit-tree right?

adam2392:
> Quick question: you're not planning on using hyppo as a dependency of scikit-tree right?

No I would say the other way around makes more sense. If we want to incorporate and move `hyppo` stuff into `pywhy-stats`, then `scikit-tree` would be an optional dependency if one uses trees to do statistical testing, otw not required.

sampan501:
Ok good. I was planning on adding it as an option dependency to hyppo

adam2392:
@sampan501 I'm guessing you might be busy at your internship, so hope it's going well!

If so, @YuxinB would you be interested in finishing this PR that Sampan started? The ability to convert an existing tree class that we have to an honest estimator will be nice to test the different trees for MI/CMI.

sampan501:
Sounds good!

PSSF23 (commented):
@adam2392 I resolved some of your comments on the code and fill in the `meson` and `init` entries. Seems like the remaining tasks are unit tests and the "prodensity tree" stuff?

adam2392:
Yep! Lmk if you run into any issues or want to discuss

PSSF23 (commented):
Ok, I tried to run tests but found a lot of problems like a honest tree that has a decision tree as its estimator? Need to fix that, and all the parameters of forest classifiers are missing, need to add those as well.

Question is: Do we want `HonestTreeClassifier` to be a tree or just an empty shell? The current code has part of both which is very misleading.

adam2392:
> Question is: Do we want `HonestTreeClassifier` to be a tree or just an empty shell? The current code has part of both which is very misleading.

I think we want it to be a meta-estimator. For that matter, we want the forest to also be a meta-estimator. So sklearn only checks if there is a parameter called `estimator`, so you prolly want some extra checks to make sure the estimator is a valid tree/forest.

See some discussions here: https://github.com/neurodata/scikit-tree/pull/57#issuecomment-1480130108

I didn't catch this when @sampan501 originally implemented the PR.

adam2392:
Lmk if you need to chat at all @PSSF23

PSSF23:
Thx for answering. I will try to figure it out first.

PSSF23 (commented):
The current error I'm stuck on is this:
```
../miniconda3/envs/ndd/lib/python3.8/site-packages/sklearn_fork/utils/estimator_checks.py:635: in check_estimator
    check(estimator)
../miniconda3/envs/ndd/lib/python3.8/site-packages/sklearn_fork/utils/estimator_checks.py:1135: in check_sample_weights_not_overwritten
    assert_allclose(sample_weight_fit, sample_weight_original, err_msg=err_msg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

actual = array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
desired = array([10.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,
        1.,  1.,  1.])
rtol = 1e-07, atol = 0.0, equal_nan = True
err_msg = 'HonestForestClassifier overwrote the original `sample_weight` given during fit'
```
Is the sample weight overwritten during honest fitting? It seems that sklearn api doesn't allow it for estimators.

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,495 @@
...
+
+        Returns
+        -------
+        self : object
+            Fitted estimator.
+        """
+        X, y = check_X_y(X, y)
+
+        # Account for bootstrapping too
+        if sample_weight is None:
```
adam2392: Is this causing the issue @PSSF23?

I don't see why it's necessary since by default None implies all are weighted equally.

adam2392:
Maybe you want to not inherit from the tree and forest base classes. 

You just need an API wrapper around the 'honesttree.estimator.<functions>' 

E.g. option 2 in the discussion I linked.

sktree/ensemble/_honest_forest.py
PSSF23: It doesn't change anything, I tried to use it to resolve the error.

PSSF23:
> Maybe you want to not inherit from the tree and forest base classes.
> 
> You just need an API wrapper around the 'honesttree.estimator.'
> 
> E.g. option 2 in the discussion I linked.

That wrapper can only be at one level of implementation, ie, tree or forest. We cannot wrap a forest of wrapped trees: honestforest(forest(honesttree(decisiontree))). Also that should not affect the estimator checks, so the sample weight problem will still be there. Unless we don't need the class to be sklearn compliant.

adam2392:
> That wrapper can only be at one level of implementation, ie, tree or forest. We cannot wrap a forest of wrapped trees: honestforest(forest(honesttree(decisiontree))). Also that should not affect the estimator checks, so the sample weight problem will still be there. Unless we don't need the class to be sklearn compliant.

Shouldn't the following work for both tree and forest?

```
class HonestTree(metaestimatorMixin):
    def init(estimator=DecisionTreeClassifier(), other_kwargs):
       ...
    
    def fit(self, X, y, sample_weight):
          ...
          self.estimator.fit(X, y, sample_weight)
          ...
   
    def apply(self, X):
         self.estimator.apply(X)
```

PSSF23:
So forest level is `HonestTree(estimator=RandomForestClassifier)`? Will that make the trees honest?

Ronan's method first makes the trees honest and then ensemble them together to make the honest forest. If we wrap around  a random forest we wouldn't get the honest indices for each tree.

adam2392:
> So forest level is `HonestTree(estimator=RandomForestClassifier)`? Will that make the trees honest?
> 
> Ronan's method first makes the trees honest and then ensemble them together to make the honest forest. If we wrap around a random forest we wouldn't get the honest indices for each tree.

My understanding (it could be wrong) is that the honest tree/forest fits on data and then re-sets the leave's vote by dropping another set of data in. In subsampled honest trees, this is a completely held out (X, y) pair. In propensity trees, this is (X, y), where the initial tree is fit using (X, T), where T stands for treatment.

So if we have:

`HonestForest(estimator=RandomForestClassifier)` shouldn't we be able to do the same thing? Fit and then re-drop the samples using either subsample, or propensity approach and refit the leaves?

PSSF23:
How about the bootstrapping part? The trees are not fitted using the same sample. 

Also regarding the sample weight part, yes, we do change it during fitting honest trees. That's how we exclude the honest samples from fitting:

        sample_weight[self.honest_indices_] = 0

So if we do not resolve this the estimator checks won't pass.

adam2392:
> How about the bootstrapping part? The trees are not fitted using the same sample.

Hmm, we could either i) keep track of the bootstrapped sample indices, which would require some change in the `_forest.py` file inside sklearn, or ii) we could just use the approach then currently here:

```
class HonestForest(BaseForest):
      def __init__(self, estimator=HonestTree(), ...):
     
      def fit(...):
             # add some extra logic here
             super().fit(X, y, sample_weight)

```

This would then allow the user to specify `estimator = HonestTree(ObliqueDecisionTree(...)`, which then fits an honest oblique forest. 

> Also regarding the sample weight part, yes, we do change it during fitting honest trees. That's how we exclude the honest samples from fitting:
> 
> ```
>     sample_weight[self.honest_indices_] = 0
> ```

Can't we just pass in the subsample of X/y (assuming the honest option chosen is subsampling and not propensity)?

`self.estimator.fit(X[self.structure_indices_,:], y[self.structure_indices_, :], sample_weight[self.structure_indices_]`, or will this not resolve the issue either?

adam2392:
Worst case scenario, we can hack what tests are expected to fail with the sample indices thing but it's not desirable. I think there is a more elegant way of fixing it that will help future proof us.

PSSF23:
ii) is already there I believe. Probably some more checks to make sure that the `estimator` param is an honest tree wrapper.

I looked over the subsample method, but that still requires a different sample weight than the one in param, right? If we only want to pass that test, probably just setting `honest_fraction` to be default 0 is enough. Yeah I want to make it more future proof as well.

adam2392:
> ii) is already there I believe. Probably some more checks to make sure that the `estimator` param is an honest tree wrapper.

Yeah and I would also test if we can pass in an "oblique honest tree".

> I looked over the subsample method, but that still requires a different sample weight than the one in param, right? If we only want to pass that test, probably just setting `honest_fraction` to be default 0 is enough. Yeah I want to make it more future proof as well.

Yeah you mean in sklearn? 

I'm wondering why not just pass in the subsampled X/y/sample_weight based on the `honest_indices`? This improves the speed of the tree fitting because it doesn't have to consider extraneous X/y.

By passing in sample_weight=0 for certain samples, it still sorts X and iterates over X/y. It is only artificially ignoring it in the Criterion class where it computes the contribution of each sample and weights it by the `sample_weight=0`.

PSSF23:
Sure, I will try some oblique tests. 

A potential problem of only fitting the subsampled dataset is: some class labels might be missing. Suppose the honest fraction takes away all the y=1, when applying those samples to set leaf nodes the classifier would break.

adam2392:
> Sure, I will try some oblique tests.
> 
> A potential problem of only fitting the subsampled dataset is: some class labels might be missing. Suppose the honest fraction takes away all the y=1, when applying those samples to set leaf nodes the classifier would break.

Can't we use Stratified splitting strategy for this for classification? (e.g. StratifiedKFold keeps the balance of `y` labels)

PSSF23:
That changes the algorithm, right? And how about when only 1 sample has that specific y label? I wonder if sklearn bootstrapping has some protection against that (not by stratifying).

PSSF23:
Overall I think the current sample_weight approach is actually robust, and it's not worth the cost (protection/stratifying) to subsample everything.

adam2392:
> Overall I think the current sample_weight approach is actually robust, and it's not worth the cost (protection/stratifying) to subsample everything.

So this sets sample_weight to 0 when fitting, but isn't this suspect to the same issue you raised? Some samples used for fitting might now end up having only `y=0`, or `y=1`. Moreover, when fitting the leaves now also has the same issue?

PSSF23:
"sets sample_weight to 0" is the direct cause of failing the sklearn estimator check (`sample_weight` changed during fitting).

Subsampling the data would both create other problems (missing y labels for some sets) and keep having the same check error, as it also changes the `sample_weight`.

adam2392:
> "sets sample_weight to 0" is the direct cause of failing the sklearn estimator check (`sample_weight` changed during fitting).
> 
> Subsampling the data would both create other problems (missing y labels for some sets) and keep having the same check error, as it also changes the `sample_weight`.

What I'm confused on is why isn't "setting sample_weight to 0" just as problematic? I can see that it may work in the cases we want, but without stratifying or raising an error when the random subsample is imbalanced, than the sample_weight=0 method is just as bad:

1. Set some sample_weight =0  and do `estimator.fit(X, y, sample_weight)` and the samples that are chosen are all `y=1`: then we have the same error. Moreover, the tree has to keep track of a larger dataset that it implicitly isn't using. So it's dinged on both performance and same error that we're discussing.
2. Based on which samples you would apply `sample_weight=0`, call this `honest_subsamp_idx` and do `estimator.fit(X[honest_subsamp_idx,:], y[honest_subsamp_idx], ...)` would give you the same answer as 1 when you don't have problematic subsampled indices and also it keeps track of a smaller dataset.

I currently am under the impression that in 2. if there is a problematic subsample, then it is also the case in 1. Moreover, option 1 has worse performance always.

PSSF23:
When the classifier fits the data, it looks over `y` and counts its unique values. Setting some sample weights to 0 doesn't change this process, as all `y` labels are still there.

For subsampled data pairs, some `y` labels could be entirely missing. To solve this, we would need to do what I did in streaming trees, giving all the labels explicitly:

https://github.com/PSSF23/scikit-learn-stream/blob/a2aab5f7aa8a67ccb3dc974adc7bb62580fe5fc2/sklearn/tree/_classes.py#L219-L242

Both approaches modify the sample weights.

adam2392:
> When the classifier fits the data, it looks over `y` and counts its unique values. Setting some sample weights to 0 doesn't change this process, as all `y` labels are still there.
> 
> For subsampled data pairs, some `y` labels could be entirely missing. To solve this, we would need to do what I did in streaming trees, giving all the labels explicitly:
 
If the only issue is that the Cython tree/splitter/criterion needs to know all possible values of `y`, I think the subsampling approach and forcing at least one sample of each class to be part of both sets is better still. Honest subsampling is supposed to be "random", so I think this procedure while breaks it a bit, prolly works in practice?

Forcing sampling weights to 0 decreases performance significantly since you have to still sort over features even if they are irrelevant for your splitting.

I mean I guess the subsampling thing is irrelevant because in the long-run we want to mainly use propensity trees for causal inference. The honest thing is prolly also irrelevant (except in theory) for uncertainty forests cuz I suspect the decrease in sample size outweighs the "unbiasedness of honesty".

If you want to implement the sample_weight thing, I would just document this is highly inefficient when sample size scales up. Then we just want this implementation to cleanly support the propensity tree, where we fit on some set of data (i.e. T) and then allow an API to reset the leaves using a different set of data (i.e. y). Lmk if this sounds good?

PSSF23:
I'll add the documentation for it. And I will add propensity api after fixing the tests, which requires some refactoring.

The "at least one" method for subsampling would still break when the structure fraction could not include all the class labels. For example, `honest_fraction = 0.5`, `y = [1,2,3,4,5]` will break the subsampling.

YuxinB:
Sorry, didn't check this email. Sure

YuxinB:
Oh sorry again, is that what Hao has done this week? Sorry again I didn't
check emails right away.

PSSF23:
@YuxinB I'm already working on it since Jovo assigned it to me on Monday.

PSSF23 (approved):
As all tests in the old honest tree repo passed and the simulation figure is replicated:
overlapping_gaussians
I believe the current patch satisfies this PR's requirements. We can merge the honesty part first and manage other utilities like propensity trees in next step.

PSSF23 (approved):
As all tests in the old honest tree repo passed and the simulation figure is replicated:
overlapping_gaussians
I believe the current patch satisfies this PR's requirements. We can merge the honesty part first and manage other utilities like propensity trees in next step.

PSSF23:
@adam2392 your thoughts?

sktree/tests/test_honest_tree.py
```diff
@@ -0,0 +1,88 @@
...
+iris = datasets.load_iris()
+rng = np.random.RandomState(1)
+perm = rng.permutation(iris.target.size)
+iris.data = iris.data[perm]
+iris.target = iris.target[perm]
+
+
+@pytest.mark.parametrize("criterion", ["gini", "entropy"])
+@pytest.mark.parametrize("max_features", [None, 2])
+def test_iris(criterion, max_features):
```
adam2392: Can we test various trees as input to `HonestTreeClassifier`? 

E.g.

HonestTreeClassifier(ObliqueDecisionTree)
HonestTreeClassifier(PatchObliqueDecisionTree)

sktree/tests/test_honest_forest.py
```diff
@@ -0,0 +1,139 @@
...
+    X[10:] *= -1
+    y = [0] * 10 + [1] * 10
+    clf = clf.fit(X, y)
+    np.testing.assert_array_equal(clf.predict(X), y)
+
+
+@pytest.mark.parametrize("criterion", ["gini", "entropy"])
+@pytest.mark.parametrize("max_features", [None, 2])
+def test_iris(criterion, max_features):
+    # Check consistency on dataset iris.
```
adam2392: Can we test various trees as input to `HonestForestClassifier`? 

E.g.

HonestTreeClassifier(ObliqueDecisionTree)
HonestTreeClassifier(PatchObliqueDecisionTree)

sktree/tree/_honest_tree.py
```diff
@@ -0,0 +1,507 @@
...
+        elif self.honest_prior == "uniform":
+            proba[zero_mask] = 1 / self.n_classes_
+        elif self.honest_prior == "ignore":
+            proba[zero_mask] = np.nan
+        else:
+            raise ValueError(f"honest_prior {self.honest_prior} not a valid input.")
+
+        return proba
+
+    def _impute_missing_classes(self, proba):
```
adam2392, line 440: Can we add a test for this function when building a tree?

sktree/tests/test_honest_tree.py
adam2392, line 1: Is it possible to add a test of the form:

```
@parametrize_with_checks(
    [
        HonestTreeClassifier(random_state=12),
        HonestTreeClassifier(random_state=12, estimator=ObliqueDecisionTreeClassifier()),
        HonestTreeClassifier(random_state=12, estimator=PatchObliqueDecisionTreeClassifier()),
    ]
)
def test_sklearn_compatible_estimator(estimator, check):
    # TODO: remove when we implement Regressor classes
    if check.func.__name__ in ["stuff expected not to pass"]:
        pytest.skip()
    check(estimator)
```

adam2392: This ensures we've handled a lot of edge cases that sklearn checker tests for

adam2392 (commented):
It mostly looks good. It would be good to tighten up the testing here while the PR is open.

Also, would it be easy/fast example to include the figure that you regenerated from the paper as an example in `examples/`? Or is it a slow running simulation.

sktree/tests/test_honest_tree.py
PSSF23: All of them won't pass. We did not solve the `sample_weight` issue yet. But sure, I can add it.

adam2392: We can add the pattern:

```
if check.func.__name__ in ["stuff expected not to pass"]:
        pytest.skip()
```

to fix that. How many tests don't pass?

sktree/tree/_honest_tree.py
PSSF23: Currently the function seems to be a failsafe for trees that don't use the whole y labels, similar to what we discussed earlier about subsampling. Right now, however, none of our tree estimators seem to need this function... I would just keep it for future proof, but I don't know how we can test it now.

adam2392: I see but implicitly with the sample weight hack this situation never occurs right? 

Let's just remove it then? I don't think we'll even need in future. 

For propensity tree the only thing we need is the refitting leaves function.

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,488 @@
...
+    search of the best split. To obtain a deterministic behaviour during
+    fitting, ``random_state`` has to be fixed.
+
+    Honesty is a feature of trees that enables unbiased estimates of confidence
+    intervals. The default implementation here is using double sampling to
+    implement honesty. The amount of samples used for learning split nodes vs
+    leaf nodes is controlled by the ``honest_fraction`` parameter. This forest
+    classifier is a "meta-estimator" because any tree model can be used in the
+    classification process, while enabling honesty separates the data used for
+    split and leaf nodes.
```
adam2392: I would also add a note here saying something like

in order to enforce honesty but also enable the tree to have access to all y labels, we set sample_weight to 0 for a random subset of samples. 

This results in inefficiency when building trees using a greedy splitter as we still sort over all values of X.

sktree/tests/test_honest_tree.py
PSSF23: Sorry I didn't understand your meanings earlier. I have added the test and skipped all classifiers.

adam2392: Sorry what I meant was the `check_estimator(est)` will run a bunch of checks. Do they all fail because of the sample_weight issue? If not, what are the ones that fail and assuming they are all related to `sample_weight`, then we can ignore them.

For example I do this in `test_tree.py`

```
    if check.func.__name__ in ["check_requires_y_none"]:
        pytest.skip()
```

because `y!=None` for our trees.

PSSF23 (commented):
It seems that I was wrong about the impute function. It seems to affect the performance under very low honest fraction, but I don't know why the coverage shows not covered.

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,493 @@
...
+    ``max_features=n_features`` and ``bootstrap=False``, if the improvement
+    of the criterion is identical for several splits enumerated during the
+    search of the best split. To obtain a deterministic behaviour during
+    fitting, ``random_state`` has to be fixed.
+
+    Honesty is a feature of trees that enables unbiased estimates of confidence
+    intervals. The default implementation here is using double sampling to
+    implement honesty. The amount of samples used for learning split nodes vs
+    leaf nodes is controlled by the ``honest_fraction`` parameter. In order to
+    enforce honesty but also enable the tree to have access to all y labels,
```
adam2392:
```suggestion
    enforce honesty, but also enable the tree to have access to all y labels,
```

sktree/ensemble/_honest_forest.py
```diff
@@ -0,0 +1,493 @@
...
+    we set sample_weight to 0 for a random subset of samples. This results in
+    inefficiency when building trees using a greedy splitter as we still sort
+    over all values of X.
```
adam2392:
```suggestion
    over all values of X. We recommend using propensity trees if you are computing causal effects.
```

sktree/tests/test_honest_tree.py
adam2392: @PSSF23 the way it's implemented rn, it skips all checks from `check_estimator`. Does it actually fail all tests? I think some tests are in fact not dependent on the `sample_weight` hack and if it fails, may be some other deeper issue that we haven't covered.

PSSF23: yes they all fail at sample weight like the error I showed earlier.

adam2392 (approved):
LGTM once CIs pass.

adam2392:
Thanks @sampan501 and @PSSF23 !

PR #64, opened 2023-04-14
[ENH] faster affinity and dissimilarity matrix computation

sampan501:
None

- Replace nested for loop in affinity calculation unsupervised forest with np.equal.outer and one for loop. Should for faster for large `n_estimators`

## Before submitting

- [x] I've read and followed all steps in the [Making a pull request](https://github.com/py-why/pywhy-graphs/blob/main/CONTRIBUTING.md#making-a-pull-request)
- [x] I've updated or added any relevant docstrings following the syntax described in the
- [x] If this PR fixes a bug, I've added a test that will fail without my fix.
- [x] If this PR adds a new feature, I've added tests that sufficiently cover my new functionality.

## After submitting


sampan501:
@adam2392 Thoughts?

Also, I believe the style checks are failing for files I did not edit

adam2392 (commented):
Per our discussion, maybe we should move this functionality to a ensemble/neighbors.py file so it can be used with supervised trees as well?

Any chance you have a short good unit test that works for this function? I realized we don't have one rn.

adam2392:
> @adam2392 Thoughts?
> 
> 
> 
> Also, I believe the style checks are failing for files I did not edit

Can you just run black on the whole repo?

sampan501:
Yep I can do that

sampan501:
@adam2392 For the supervised versions, the options are to edit `ForestClassifier` in `scikit-learn-fork`, or just add the similarity matrix computation to each respective forest type (and the same when I write the dissimilarity matrice functions). Which do you prefer?

adam2392:
> @adam2392 For the supervised versions, the options are to edit `ForestClassifier` in `scikit-learn-fork`, or just add the similarity matrix computation to each respective forest type (and the same when I write the dissimilarity matrice functions). Which do you prefer?

So, based on our email, we want to add this functionality to all classes, but also we might want to keep the fork lightweight unless necessary. 

I would just add a Mixin class that has this functionality and add the Mixin class to all the trees/forests applicable in scikit-tree. 

We should/could also expose a functional interface, so just

```
def compute_similarity_matrix_forest(...)


def compute_dissimilarity_matrix_forest(...)
```

Then this is called in the tree/forest class methods. I think the functional approach makes more sense cuz it leans towards not adding to the tree classes API.

adam2392 (commented):
Do you mind adding the dissimilarity matrix while you're at it and the reference that Jovo mentioned in the email thread?

https://arxiv.org/pdf/1812.00029.pdf

Otw, LGTM

adam2392:
To fix the CI issues:

- run `poetry run poe format` which applies all formatting checks. You can run `poetry run poe lint` to check any flake8 issues and pydocstyle stuff.
- add `sktree/ensemble/_neighbors.py` to the corresponding `meson.build` file inside `ensemble/` to register the additional Python file.

sktree/ensemble/_unsupervised_forest.py
```diff
@@ -26,6 +26,7 @@
 from sklearn_fork.utils.validation import _check_sample_weight, check_is_fitted, check_random_state
 
 from sktree.tree import UnsupervisedDecisionTree, UnsupervisedObliqueDecisionTree
+from sktree._neighbors import _sim_matrix
```
adam2392:
```suggestion
from ._neighbors import _sim_matrix
```

sktree/ensemble/_neighbors.py
```diff
@@ -0,0 +1,58 @@
...
+
+        Parameters
+        ----------
+        X : ndarray of shape (n_samples, n_estimators)
+            For each datapoint x in X and for each tree in the forest,
+            is the index of the leaf x ends up in.
+
+        Returns
+        -------
+        prox_matrix : array-like of shape (n_samples, n_samples)
```
adam2392: Add descriptions to handle `numpydoc` errors.

sktree/ensemble/_supervised_forest.py
```diff
@@ -334,8 +335,42 @@ def __init__(
...
+        y : array-like of shape (n_samples,) or (n_samples, n_outputs)
+            The target values (class labels in classification, real numbers in
+            regression).
+        sample_weight : array-like of shape (n_samples,), default=None
+            Sample weights. If None, then samples are equally weighted. Splits
+            that would create child nodes with net zero or negative weight are
+            ignored while searching for a split in each node. In the case of
+            classification, splits are also ignored if they would result in any
+            single class carrying a negative weight in either child node.
+        Returns
```
adam2392:
```suggestion

        Returns
```

sktree/ensemble/_supervised_forest.py
```diff
@@ -669,3 +704,37 @@ def __init__(
...
```
adam2392:
```suggestion

        Returns
```

adam2392:
I fixed the numpydoc errors, sphinx errors and also refactored this.

I think storing a dissimilarity matrix is not needed and adds just RAM cost for no gain. Someone can just take `1 - forest.similarity_matrix_` and that gives the same computation.

I also factored out the similarity computation as a function, so it can be easily used with any BaseForest method that implements `apply`.

https://github.com/neurodata/scikit-tree/pull/64/commits/f4f977ecb4b49df0a24e1806b49d1425db90a737

adam2392 (approved):
All CIs fixed, and I applied some fixes. LGTM

sampan501:
Looks good to me!

adam2392:
@sampan501 I removed the automatic computation of similarity matrix as this caused unnecessary RAM/CPU usage on the docs building. I think we can enable it in the future if necessary, but one can just call `compute_similarity_matrix(X)`.

adam2392:
@sampan501 I noticed we never added `dissimilarity_matrix_`. What is the way to compute this that's in-line w/ the affinity matrix?

Moreover, are we missing a division by the `affinity_matrix.max()` somewhere?

sampan501:
I think we talked about not storing both in memory since dissim is just 1-sim.

Also since the max is 1, there is no need for dividing by max

Issue #46, opened 2023-03-06
[ENH] Add "extratrees" (i.e. randomized trees) for oblique trees


adam2392:
Add random splitter for oblique trees into the codebase to allow "ExtraObliqueTreesClassifier".

This will improve runtime and also act as an analogous improvement to the oblique trees similar to how `extratreesclassifier` generalizes `decisiontreeClassifier` in scikit-learn

xref: https://link.springer.com/article/10.1007/s10994-006-6226-1

adam2392:
This is complementary to binning #25

adam2392:
Someone would basically have to look at the oblique splitter code (this implements "best oblique splitter") currently in scikit-tree, and then the random splitter code in scikit-learn and add a corresponding set of Cython code that implements "oblique random splitter"

SUKI-O:
Am I totally off-base in thinking I need to implement an `extra` version of `class BaseObliqueSplitter(Splitter)`? Or higher-level ` class BestObliqueSplitter(ObliqueSplitter):` ? We want to pick the partition threshold randomly for all features (weighted by the weights), unlike  the `class RandomSplitter` in schikit-learn picks a feature randomly AND picks threshold within the feature randomly.  We want that latter part implemented when picking the split in ObliqueSplitter?

adam2392:
> Am I totally off-base in thinking I need to implement an `extra` version of `class BaseObliqueSplitter(Splitter)`? Or higher-level ` class BestObliqueSplitter(ObliqueSplitter):` ? We want to pick the partition threshold randomly for all features (weighted by the weights), unlike the `class RandomSplitter` in schikit-learn picks a feature randomly AND picks threshold within the feature randomly. We want that latter part implemented when picking the split in ObliqueSplitter?

I would initially guess implement a subclass of `BaseObliqueSplitter` called `RandomObliqueSplitter`, which picks projection matrix and threshold randomly.

SUKI-O:
from `BaseObliqueSplitter` line 264-274
```
 if current_proxy_improvement > best_proxy_improvement:
                        best_proxy_improvement = current_proxy_improvement
                        # sum of halves is used to avoid infinite value
                        current_split.threshold = feature_values[p - 1] / 2.0 + feature_values[p] / 2.0

                        if (
                            (current_split.threshold == feature_values[p]) or
                            (current_split.threshold == INFINITY) or
                            (current_split.threshold == -INFINITY)
                        ):
[...]

PR #75, opened 2023-05-20
[ENH] Add "extratrees" for oblique trees #46

SUKI-O:
Implement #46 

- Add 'Random' threshold selection for ObliqueTrees 

## Before submitting

## After submitting


SUKI-O:
In Scikit-Learn `ExtraTree`, `node_split_random` function selects a feature randomly and selects a threshold randomly. 
```        # Draw a random threshold
        current_split.threshold = rand_uniform(
            min_feature_value,
            max_feature_value,
            random_state,
        )
```
It's picking rand_uniform between (min feature val, max feature val).  But why can we sample `p` from `range(p)` for example? Wouldn't that be a better sampling strategy? The latter is more cognizant of the sample distribution at that node, no?

adam2392 (changes requested):
Thanks! Good first try. This doesn't work the way intended as of now:

- If you look at the design of the trees/splitter/criterion and Python wrapper, the Python wrapper instantiates each class and then passes them to some Tree Builder. The tree builder calls the API of each class. In https://github.com/neurodata/scikit-learn/blob/a799dbe5feb1a0956c24b10ac1ea27f1eadfb76f/sklearn_fork/tree/_tree.pyx#L505, we see that the way it is called is in line w/ the abstract API defined by `BaseSplitter.node_split`. The current implementation would never be able to be used.
- The only difference between extra trees and normal trees is that instead of sorting and choosing the best threshold at each split point, a random threshold is chosen. So: https://github.com/neurodata/scikit-learn/blob/a799dbe5feb1a0956c24b10ac1ea27f1eadfb76f/sklearn_fork/tree/_splitter.pyx#L714-L722 vs https://github.com/neurodata/scikit-learn/blob/a799dbe5feb1a0956c24b10ac1ea27f1eadfb76f/sklearn_fork/tree/_splitter.pyx#L409. Other lOC might be necessary due to the minor change in logic. I.e. we don't need to sort.
- Re above point, rather than evaluate all splits, we don't need to do that anymore. A split point is just randomly chosen. I.e. we don't need to evaluate all split points. https://github.com/neurodata/scikit-learn/blob/a799dbe5feb1a0956c24b10ac1ea27f1eadfb76f/sklearn_fork/tree/_splitter.pyx#L427-L462.

Moreover, the logic is slightly different in the oblique splitter implementation. So you're basically taking the oblique splitter implementation and changing the analogous logic. Lmk if this makes sense. If it helps you when speccing out the high-level changes needed, then you can also write an outline of a code plan, and I can lyk if it's in the right direction.

This requires a bit of reading the code and perhaps writing down how things are moving and such to fully understand the tree implementation.

SUKI-O:
At a high level, I was thinking of the following:
- When `ExtraObliqueTree` is initialized, its `splitter` is set to `random` which instantiates `RandomObliqueSplitter` 
- It inherits `BestObliqueSplitter` but overwrites the  `node_split` function.  
- The `RandomObliqueSplitter.node_split` method picks features via `sample_proj_mat` as in `BestObliqueSplitter` and iterates over each chosen feature except that it chooses threshold `p` randomly.
     - Check if the `min_samples_leaf` criterion is met, else pick another random threshold 
     - Check if the `min_weight_leaf` criterion is met, else pick another random threshold 
- Reorganize into samples[start:`p`] + samples[`p`:end] --- same as in `BaseObliqueSplitter`

Am I totally off? 
Don't we need to sort it anyway in order to split the samples into children nodes? I guess you can do it without sorting but computationally no less expensive, I don't think.   
And to check for the `min_weight_leaf` criterion, we still need to evaluate at that particular choice of threshold, right?
Thanks again for the help :)

adam2392:
> At a high level, I was thinking of the following:
> 
> * When `ExtraObliqueTree` is initialized, its `splitter` is set to `random` which instantiates `RandomObliqueSplitter`
> * It inherits `BestObliqueSplitter` but overwrites the  `node_split` function.
> * The `RandomObliqueSplitter.node_split` method picks features via `sample_proj_mat` as in `BestObliqueSplitter` and iterates over each chosen feature except that it chooses threshold `p` randomly.
>   
>   * Check if the `min_samples_leaf` criterion is met, else pick another random threshold
>   * Check if the `min_weight_leaf` criterion is met, else pick another random threshold
> * Reorganize into samples[start:`p`] + samples[`p`:end] --- same as in `BaseObliqueSplitter`

Yeah more or less. 

And then some FYI, in case it isn't clear in the code: `start, p, end` are pointers to the sample array. I.e. if you have 100 samples, these are pointers to indices to prevent data copying. `start:p` = left child usually and `p:end` = right child usually and `start:end` = current node. 

> Am I totally off? Don't we need to sort it anyway in order to split the samples into children nodes? I guess you can do it without sorting but computationally no less expensive, I don't think. 

No this is more or less in  the right direction. The whole point of choosing a threshold at random is so you can i) avoid sorting and ii) avoid computing the criterion for each possible split threshold.

> And to check for the `min_weight_leaf` criterion, we still need to evaluate at that particular choice of threshold, right? 

Well, more so that once we pick the threshold yes we have to evaluate it. But see comment above. The main point of random is to prevent the need to evaluate all particular choices of thresholds. This complements the randomness induced by: i) random subsetting of data is fed into each tree, ii) random features are picked for possible splitting on at each node in trees.

adam2392 (commented):
Copying / pasting your comment in Slack here to leave comments (btw feel free to just comment on GH; it is a lot easier to respond):

> I created RandomObliqueSplitter that inherits from BestObliqueSplitter which has sample_proj_mat function. It overwrites the node_split function with 'random' version that picks current_split.threshold at random.

Scikit-learn's random splitter has a slightly different design, but the high-level logic is more or less similar. I would recommend reviewing this function and writing out the logic for yourself on paper.

https://github.com/scikit-learn/scikit-learn/blob/4f89e717570e76650ca6f73e25c5151f8902dad5/sklearn/tree/_splitter.pyx#L625

> It ensures min_weight_leaf and min_samples_leaf criteria are met but besides that it does not sort or evaluate other criterion, just once for the randomly chosen threshold.  It builds but many test fails and I am clearly not understanding the root cause. Is there some specific consideration needed for testing 'random' version or should all the tests be applicable regardless?

Currently, you are sorting still, so I don't think this is accurate.

I would take a look at the sklearn implementation of best splitter and random splitter and write down the logic to compare. Next I would take a look at our oblique best splitter.

sktree/tree/_oblique_splitter.pyx
```diff
@@ -453,3 +477,198 @@ cdef class BestObliqueSplitter(ObliqueSplitter):
...
+                start,
+                end,
+                samples,
+                feature_values,
+                &self.proj_mat_weights[feat_i],
+                &self.proj_mat_indices[feat_i]
+            )
+
+            # Sort the samples
+            sort(&feature_values[start], &samples[start], end - start)
```
adam2392: See here you are still sorting.

SUKI-O: # Draw a random threshold
        current_split.threshold = rand_uniform(
            min_feature_value,
            max_feature_value,
            random_state,
        )

In `node_split_random` in `Splitter`, it samples a random (uniform) threshold between `min_feature_value` and `max_feature_value` as above and find and set `.pos` according to it by `partitioner. partition_samples. To find `min_feature_value` and `max_feature_value`, it uses `partitioner.find_min_max`, which runs iteratively each feature value and is basically the same as sorting.  I was going to use that to get `min_feature_value` and `max_feature_value` then follow the same logic as `node_split_random`. But I had a question: can we just randomly sample `current_split.pos` from `rand_int(start+min_samples_leaf, end, &self.rand_r_state)`? Is that a bad sampling strategy?  Is either better than the other?  If the latter is an acceptable sampling method, it'll be faster, (and no sorting).

adam2392: I'm not sure I understand the question. You wouldn't randomly sample `pos`. Also finding min/max is not the same as sorting. Sorting is O(nlog(n)). Finding min/max is O(n). 

The pseudocode for RandomSplitter.node_split is the following. The only difference is that `compute_features()` computes a projection of the data, rather than just choosing a column of the data. I think sklearn's random splitter does some extra stuff like keeping track of what columns contain constant values, whereas we don't do that.

```
# Some checks and stuff are missing, but this is the general gist, which is what's in sklearn's random splitting
for idx in range(max_features):
    # for each projection, compute the features over all samples
    # (n_samples,) vector
    current_feature_values = compute_features_over_samples

    # find min and max of current feature values
    min, max = find_min_max(current_feature_values)

    # draw random thresohld
    current_split.threshodl = rand_uniform(min, max, randstate)

    # partition samples given the threshold by moving feature values and samples inplace
    current_split.pos = partition_samples()

    # evaluate split
    criterion.reset()
    criterion.update(current_split.pos)

    current_proxy_improvement = self.criterion.proxy_impurity_improvement()
    if current_proxy_improvement > best_proxy_improvement:
        best_proxy_improvement = current_proxy_improvement
        best = current  # copy
    
    # reorganize samples inplace samples[start:best.pos] + samples[best.pos:end]
    # copy of the previous code
```

Possibly some good papers on this topic are the extra trees publication linked in the GH issue and the original random forest paper by Breiman.

adam2392:
Hi @SUKI-O lmk when you feel this is ready for review. It might help if you are able to write a short example using simulated data to show how this is faster than the normal oblique best splitter when building an Oblique forest.

It should also ideally be similar in performance.

Some example for inspiration:

- https://scikit-learn.org/stable/auto_examples/ensemble/plot_forest_iris.html#sphx-glr-auto-examples-ensemble-plot-forest-iris-py

SUKI-O:
> Hi @SUKI-O lmk when you feel this is ready for review. It might help if you are able to write a short example using simulated data to show how this is faster than the normal oblique best splitter when building an Oblique forest.
> 
> It should also ideally be similar in performance.
> 
> Some example for inspiration:
> 
> * https://scikit-learn.org/stable/auto_examples/ensemble/plot_forest_iris.html#sphx-glr-auto-examples-ensemble-plot-forest-iris-py

Do you think I can/should use the dataset used for ObliqueRF like [this](https://github.com/neurodata/scikit-tree/blob/d179de3305cd37520b419ed4f012b7bf741d3882/examples/plot_oblique_random_forest.py#L38C1-L53C24) to run the test/example? If not, any suggestions on what may work best for this application?

adam2392:
Sure you can use that dataset. 

I would also have an example comparing the decision boundaries on the iris dataset.

examples/plot_extra_oblique_random_forest.py
adam2392, line 1: This shouldn't just be an exact copy of the other example, else we would just compress these into one example.

Perhaps add some comments about what the performance is relative to the other example, or explicitly compute it. Also, if the goal is comparisons, then only one dataset is needed to be shown.

sktree/tree/_oblique_splitter.pyx
```diff
@@ -31,6 +31,30 @@ cdef inline void _init_split(ObliqueSplitRecord* self, SIZE_t start_pos) noexcep
     self.threshold = 0.
     self.improvement = -INFINITY
 
+
+cdef INT32_t find_index(DTYPE_t[:] feature_values, DOUBLE_t threshold) noexcept nogil:
```
adam2392: Is this used?

sktree/tree/_oblique_splitter.pyx
```diff
@@ -596,3 +620,205 @@ cdef class BestObliqueSplitter(ObliqueSplitter):
...
+                samples,
+                feature_values,
+                &self.proj_mat_weights[feat_i],
+                &self.proj_mat_indices[feat_i]
+            )
+
+            # find min, max of the feature_values
+            self.find_min_max(feature_values, &min_feature_value, &max_feature_value)
+
+            # Draw a random threshold
```
adam2392, line 583:
```suggestion
            # XXX: we can still add some logic here to keep track of constant features if they exist
            # Draw a random threshold
```

docs/whats_new/v0.1.rst
```diff
@@ -35,7 +35,7 @@ Changelog
 - |Feature| Implementation of HonestTreeClassifier, HonestForestClassifier, by `Sambit Panda`_, `Adam Li`_, `Ronan Perry`_ and `Haoyin Xu`_ (:pr:`57`)
 - |Feature| Implementation of (conditional) mutual information estimation via unsupervised tree models and added NearestNeighborsMetaEstimator by `Adam Li`_ (:pr:`83`)
 - |Feature| Add multi-output support to HonestTreeClassifier, HonestForestClassifier, by `Ronan Perry`_, `Haoyin Xu`_ and `Adam Li`_ (:pr:`86`)
-
+- |Feature| Implementation of  ExtraObliqueDecisionTreeClassifier, ExtraObliqueDecisionTreeRegressor by `SUKI-O`_ (:pr:`75`)
```
adam2392: We are now on v0.2 development, so I would add this to the `v0.2.rst` file instead.

adam2392 (commented):
The Cython code looks more correct to me now, but make sure you remove any extraneous code that is not being used.

In the examples, these are user-facing documentation that we want to hold to a pretty high standard because it should convey some "concept". In your case, it is comparing "extratrees*" vs "normal trees*" in some classification dataset. 

We want to see how much time is saved vs how much performance changes. 

Holding these examples to a high standard also helps you because if you ever wonder how to run something, you can refer back to the example and have high-quality sample code that you previously wrote.

sktree/tree/_oblique_splitter.pyx
SUKI-O: In the sklearn version of Splitter class, the features are organized so constants are at the front and non-constants are at the back. But does it make sense to do this in Oblique trees? My guess was no, in which case, how do we use the variable `n_constant_features`?

examples/plot_extra_oblique_random_forest.py
SUKI-O: It currently outputs the time taken for each dataset like the below:
```
                        time_taken
dataset          model            
PhishingWebsites EORF     4.294050
                 ORF      5.367431
cnae-9           EORF    32.427807
                 ORF     31.753967
wdbc             EORF     0.372672
                 ORF      0.981676
```
Also added a boundary plot on iris. But for it uses only two features each as it needs to be plotted in 2D. I guess that's ok.

sktree/tree/_oblique_splitter.pyx
adam2392: We don't use it rn but in the future we can.

examples/plot_extra_oblique_random_forest.py
adam2392: Hmm that's weird. It should run significantly faster on higher-dimensional data. What is `max_features` and `max_depth` and the actual tree depths?

I'm guessing `cnae-9` is the highest-sample size and highest-dimensional dataset since it took like 10x longer than the rest. I'm surprised then the extra-trees version actually performs slower.

sktree/tree/_oblique_splitter.pyx
adam2392: This is just to add an in-line comment, so we remember to do so in the future.

Basically if we run OF and at any point we randomly choose only one feature and realize it is a constant feature, we can note that.

examples/plot_extra_oblique_random_forest.py
```diff
@@ -0,0 +1,142 @@
+"""
+===============================================================================
+Plot extra oblique forest and oblique random forest predictions on cc18 datasets
+===============================================================================
+
+A performance comparison between extra oblique forest and standard oblique random
+forest using three datasets from OpenML benchmarking suites.
+
+Extra oblique forest uses Extra-trees as base model which differ from classic
```
adam2392: Extra oblique forests uses extra oblique trees as base model, not extra-trees. Extra-trees in scikit-learn corresponds to axis-aligned splits, whereas oblique trees correspond to oblique splits (sparse linear combinations of features)

examples/plot_extra_oblique_random_forest.py
SUKI-O: `max_features=None` and `max_depth=None`.  As for the `cnae-9` dataset, it's 100x856 so, `max_features=n_features=856` and `max_depth=sqrt(856)=~29` 
I do not like that EORF is slower than ORF on `cnae-9` and pondered why.  Looking at the `cnae-9` I realized that it's super sparse data: it's binary and mostly 0s.  I feel like an Oblique tree may not work well on these kinda data?  Should I be looking for more "dense" data, do you think? LMK.

adam2392: What's the depth of the trees in OF case?

But yeah perhaps try a few different datasets to see one that does faster and we can use this as an example that does slower (assuming it's not a bug).

SUKI-O: I did a few runs on different datasets but ExtraORF does not perform much faster on wide data. Sometimes it's even slower than OF.  Not sure what I'm doing wrong. Any suggestion?

examples/plot_extra_oblique_random_forest.py
```diff
@@ -0,0 +1,329 @@
...
+# Extra Oblique Tree runs faster compared to the standard Oblique Tree,
+# while the performance is comparable or better in some cases that are
+# tested.
+
+"""
+====================================================================
+Plot the decision surfaces of ensembles of trees on the iris dataset
+====================================================================
+
+Plot the decision surfaces of forests of randomized trees trained on pairs of
```
adam2392: Please have these as separate example files.

examples/plot_extra_oblique_random_forest.py
adam2392: A few comments that would help reviewing:

- remove any extraneous code that's not being used/touched
- organize a separate file or a https://gist.github.com even that shows what you ran and a screenshot or figure that shows the result.
- include relevant metadata (i.e. depth of the trained trees)

> What's the depth of the trees in OF case?

The answer to this question would be relevant. Does this make sense why?

SUKI-O: When I run [ExtraObliqueDT vs ObliqueDT](https://gist.github.com/SUKI-O/ce2e0510560eadb3a1367534a705e91f)

The EODT runs much faster as expected`
dataset | model | score | max_depth | time_taken
-- | -- | -- | -- | --
cnae 9 | ExtraObliqueDecisionTree | 0.810000 | 72.0 | 0.030214
cnae 9 | ObliqueDecisionTree | 0.824074 | 51.0 | 0.450707

`
However, when I run [ExtraObliqueRF vs ObliqueRF with n_estimator 1 ](https://gist.github.com/SUKI-O/b24ac87f35c95179391c3bf309201787), the time diff advantage goes away though the performance improves somewhat. 

`

dataset | model | score | avg_tree_depth | time_taken
-- | -- | -- | -- | --
cnae-9 | ExtraObliqueRandomForestClassifier | 0.759259 | 54.0 | 21.0
cnae-9 | ObliqueRandomForestClassifier | 0.712963 | 34.0 | 19.0

`
I thought that (RF with n_estimator=1) <==> DT. Why is that? I am thinking this is why EORF becomes slower than ORF on some datasets. LMK what you think.

adam2392: Your gist links are messed up. Perhaps the link you pasted is incorrect?

Re: Running a forest w/ 1 tree is meaningless, so I wouldn't do that. No worries on that. As long as each tree is faster on average, then when we compute a forest it should be "significantly" faster.

The reason it takes longer seems to be the trees get trained significantly deeper. I would try fixing the max_depth to a number in both cases and see if the speed goes up even more on a different dataset.

So it seems that on the cnae-9 dataset there is 10x speedup. However, the normal random forest already is pretty fast. Can you print some details on the dataset? E.g. shape, size, type of features, etc.

adam2392: Perhaps we can run two datasets with differing characteristics. I would look around openml and play around w/ the different datasets to see.

https://www.openml.org/search?type=data&sort=runs&id=1468&status=active seems to suggest most of the features are categorical.

Perhaps you can run it on something w/ continuous features

SUKI-O: I ran some experiments to [Compare EORF vs ORFs on OpenML dataset](https://gist.github.com/SUKI-O/a002038d9b02ae047430e3fe2093e8bc). 
As you noted that ExtraRF grows much deeper than ORF and it takes longer if you set max_depth=None. When I cap the max_depth, ExtraRF runs faster in general, though the score seems to suffer a little. 
Tried both categorical and continuous data but I am not sure if that mattered as much. LMK your thoughts?

adam2392: Hmm very weird that the difference in runtime isn't more. Can you set `n_jobs=-1`, so that all the processors are used per training of forest?

I would just re-run: cnae-9 and ova_uterus.

SUKI-O: [Compare EORF vs ORFs on OpenML dataset](https://gist.github.com/SUKI-O/a002038d9b02ae047430e3fe2093e8bc) 
Ran w/ n_jobs = -1. On OVA_Uterus, ExtraRF is slower. I set the max_depth=10. 
The weird thing is that when I ran ExtraObliqueDT vs ObliqueDT, Extra one runs 10x faster. But somehow when used in RF, the advantage evaporates. Why do you is does this happen?
`
dataset | dim | model | max_depth | n_estimators | score | time_taken | avg_tree_depth 
-- | -- | -- | -- | -- | -- | -- | --
OVA_Uterus | (1545, 10935) | ExtraObliqueRandomForestClassifier | 10 | 50 | 0.929032 | 1981.0 | 10.00 
OVA_Uterus | (1545, 10935) | ObliqueRandomForestClassifier | 10 | 50 | 0.941935 | 1720.0 | 9.98
cnae-9 | (1080, 856) | ExtraObliqueRandomForestClassifier | 10 | 50 | 0.907407 | 102.0 | 10.00 
cnae-9 | (1080, 856) | ObliqueRandomForestClassifier | 10 | 50 | 0.879630 | 116.0 | 10.00 
`

adam2392: Hmmm I see. Can you see if there is a dataset with a lot of rows (high sample size) and a decent number of features?

E.g. anywhere from 10s-1000s features, but at least 10k samples?

Motivation: The speedup should come from not having to sort samples. Sorting is O(n logn), whereas if we don't sort, then it should be more like O(n)ish...

SUKI-O: Two datasets that seemed to fit the criteria
- har (10299, 561) 
- CIFAR_10_small (20000, 3072) 

`
dataset | dim | model | max_depth | n_estimators | score | time_taken | avg_tree_depth
-- | -- | -- | -- | -- | -- | -- | -- 
har | (10299, 561) | ExtraObliqueRandomForestClassifier | 10 | 50 | 0.939806 | 358.0 | 10.0 
har | (10299, 561) | ObliqueRandomForestClassifier | 10 | 50 | 0.965049 | 616.0 | 10.0 
CIFAR_10_small | (20000, 3072) | ExtraObliqueRandomForestClassifier | 10 | 50 | 0.379000 | 3492.0 | 10.0 
CIFAR_10_small | (20000, 3072) | ObliqueRandomForestClassifier | 10 | 50 | 0.399000 | 5184.0 | 10.0 

`

The time advantage is slightly better... but not like 10x. Does this make sense?

adam2392: Yes makes sense. Maybe 2x better. Can you try without limiting max_depth and converting time taken to seconds?

adam2392:
Assuming https://github.com/neurodata/scikit-tree/pull/75#discussion_r1285221550 looks good (which it seems like it will), this is mostly good to clean up. 

I would then:

- remove extraneous code that is unnecessary 
- clean up the examples a bit and review wording

examples/plot_extra_oblique_random_forest.py
SUKI-O: With max_depth=None, ExtraObliqueRF gets deeper than ObliqueRF. With `har` dataset, it doesn't seem to affect on time but with `CIFAR_10_small` ExtraObliqueRF ends up slower. Noting that they both perform poorly. This is just my personal observation but it looks like when dealing with "hard" dataset, i.e. accuracy is low, ExtraObliqueRF struggles and often worse than ObliqueRF. 
`
dataset | dim | model | max_depth | n_estimators | score | time_taken | avg_tree_depth 
-- | -- | -- | -- | -- | -- | -- | --  
har | (10299, 561) | ExtraObliqueRandomForestClassifier | None | 50 | 0.958252 | 0.899949 | 26.40 
har | (10299, 561) | ObliqueRandomForestClassifier | None | 50 | 0.967961 | 0.897108 | 20.46 
CIFAR_10_small | (20000, 3072) | ExtraObliqueRandomForestClassifier | None | 50 | 0.390500 | 30.682241 | 33.50 
CIFAR_10_small | (20000, 3072) | ObliqueRandomForestClassifier | None | 50 | 0.408000 | 26.257330 | 31.48

`

adam2392:
The loss in accuracy is not important as long as the computation is sped up. The times all do not show anything informative because it’s shown with units as days. Sent from IPhoneOn Aug 11, 2023, at 9:21 PM, SUKI ***@***.***> wrote:﻿
@SUKI-O commented on this pull request.



On examples/plot_extra_oblique_random_forest.py:
With max_depth=None, ExtraObliqueRF gets deeper than ObliqueRF. With har dataset, it doesn't seem to affect on time but with CIFAR_10_small ExtraObliqueRF ends up slower. Noting that they both perform poorly. This is just my personal observation but it looks like when dealing with "hard" dataset, i.e. accuracy is low, ExtraObliqueRF struggles and often worse than ObliqueRF.
`



dataset
dim
model
max_depth
n_estimators
score
time_taken
avg_tree_depth




har
(10299, 561)
ExtraObliqueRandomForestClassifier
None
50
0.958252
0 days 00:00:00
26.40


har
(10299, 561)
ObliqueRandomForestClassifier
None
50
0.967961
0 days 00:00:00.001000
20.46


CIFAR_10_small
(20000, 3072)
ExtraObliqueRandomForestClassifier
None
50
0.390500
0 days 00:00:00.031000
33.50


CIFAR_10_small
(20000, 3072)
ObliqueRandomForestClassifier
None
50
0.408000
0 days 00:00:00.026000
31.48



`

examples/plot_extra_oblique_random_forest.py
SUKI-O: sorry. fixed the time --> total sec.

adam2392: Is max_features set to use all the features or is it default of sqrt?

Perhaps all would help? Otw maybe we can go back to fixing the max_depth and leave it at that.

sktree/tree/_oblique_splitter.pyx
```diff
@@ -596,3 +597,240 @@ cdef class BestObliqueSplitter(ObliqueSplitter):
...
+            )
+
+            # find min, max of the feature_values
+            self.find_min_max(feature_values, &min_feature_value, &max_feature_value)
+
+            # XXX: Add logic to keep track of constant features if they exist
+            if max_feature_value <= min_feature_value + FEATURE_THRESHOLD:
+                n_found_constants += 1
+                n_total_constants += 1
+                continue
```
adam2392: I'm not sure why these lines were added, but this is the reason for the performance regression. All unit-tests do not pass w/ this.

adam2392:
@SUKI-O 

I merged in changes from main and also fixed some unnecessary LOC that were added (https://github.com/neurodata/scikit-tree/pull/75#discussion_r1296178194). If you want to re-run the exps. I suspect they will improve.

sktree/tree/_oblique_splitter.pyx
SUKI-O: I originally thought tracking constant feature was a necessary part of the extratree.. looks like you removed this already.

adam2392: No worries. The constant features is able to be easily tracked in axis-aligned trees, but for oblique trees, this is a bit more difficult, but we would like to have it.

If not implemented correctly, it induces performance regression.

xref: #105

SUKI-O:
> @SUKI-O
> 
> I merged in changes from main and also fixed some unnecessary LOC that were added ([#75 (comment)](https://github.com/neurodata/scikit-tree/pull/75#discussion_r1296178194)). If you want to re-run the exps. I suspect they will improve.

I tried to run the docs and I got the following error:
```
Unexpected failing examples:
/Users/itsukiogihara/Projects/JHU/scikit-tree/examples/plot_projection_matrices.py failed leaving traceback:
Traceback (most recent call last):
  File "/Users/itsukiogihara/Projects/JHU/scikit-tree/examples/plot_projection_matrices.py", line 84, in <module>
    splitter = BestPatchSplitterTester(
  File "_morf_splitter.pyx", line 36, in sktree.tree.manifold._morf_splitter.PatchSplitter.__cinit__
ValueError: Buffer dtype mismatch, expected 'uint8_t' but got 'long'
```
Not sure what that even means. Any idea?

adam2392:
> @SUKI-O
> 
> I merged in changes from main and also fixed some unnecessary LOC that were added ([#75 (comment)](https://github.com/neurodata/scikit-tree/pull/75#discussion_r1296178194)). If you want to re-run the exps. I suspect they will improve.

I think I fixed most of the CI errors in https://github.com/neurodata/scikit-tree/pull/75/commits/338428a1c3b443ecdc335b0cd775efa032c4e716. So feel free to just rebuild and run the exps and if we see an improvement that we hypothesize, we can proceed.

adam2392:
https://output.circle-artifacts.com/output/job/80903942-02eb-4fcd-a7ca-0d25ec870888/artifacts/0/dev/auto_examples/plot_extra_oblique_random_forest.html#sphx-glr-auto-examples-plot-extra-oblique-random-forest-py

Shows EORF is slower. I think this could be due to the really low sample size you force in the cnae-9, har, lsvt datasets. Perhaps you can increase the sample size and run it. 

You can build and run it locally to get something that makes sense.

SUKI-O:
Screenshot 2023-09-04 at 7 04 29 AM 

I've tweeked around to get better results but for wide datasets like cnae-9 and lsvt,  EORF is always slower. Does this make sense to you? 

[Edit]
Screenshot 2023-09-04 at 7 47 40 AM 

If I set max_depth = 3, max_features = 30, I could get to the above result.  Is this legit?

adam2392:
> Screenshot 2023-09-04 at 7 04 29 AM
> I've tweeked around to get better results but for wide datasets like cnae-9 and lsvt, EORF is always slower. Does this make sense to you?
> 
> [Edit] Screenshot 2023-09-04 at 7 47 40 AM
> 
> If I set max_depth = 3, max_features = 30, I could get to the above result. Is this legit?

Yes these make sense. I would add a [note](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html) explaining that you set the `max_depth=3` purely for demonstration purposes. In general, the max_depth can be tuned.  The speedup is relative to sample size. O(n) in extra trees vs O(n log(n)) in normal trees. So limiting the sample size, we might see other factors play into the speed, which explains the fact that normal trees were actually faster (e.g. normal trees might've trained to smaller depths). 

I would read the extra trees paper if that helps to communicate the key points: bias-variance-tradeoff, computational runtime. This is a good SO post too on the topic: https://stats.stackexchange.com/questions/175523/difference-between-random-forest-and-extremely-randomized-trees

examples/plot_decision_surface_iris.py
```diff
@@ -0,0 +1,174 @@
+"""
+====================================================================
+Plot the decision surfaces of ensembles of trees on the iris dataset
```
adam2392:
```suggestion
Plot the decision surfaces of ensembles of oblique trees on the iris dataset
```

examples/plot_decision_surface_iris.py
```diff
@@ -0,0 +1,174 @@
...
+====================================================================
+
+Plot the decision surfaces of forests of randomized trees trained on pairs of
```
adam2392:
```suggestion
Plot the decision surfaces of forests of randomized oblique trees trained on pairs of
```

examples/plot_decision_surface_iris.py
```diff
@@ -0,0 +1,174 @@
...
+features of the iris dataset.
+
+This plot compares the decision surfaces learned by a decision tree classifier
+(first column), by a oblique decision tree classifier (second column), and by an
```
adam2392: `plot_iris_dtc.py` already compared DTC and Oblique DTC

examples/plot_decision_surface_iris.py
adam2392, line 1: This example should only compare oblique dtc and oblique extra tree. `plot_iris_dtc.py` already compared dtc and oblique dtc.

examples/plot_extra_oblique_random_forest.py
```diff
@@ -0,0 +1,170 @@
+"""
+================================================================================
+Plot extra oblique forest and oblique random forest predictions on cc18 datasets
```
adam2392:
```suggestion
Compare extra oblique forest and oblique random forest predictions on cc18 datasets
```

sktree/tree/_utils.pyx
```diff
@@ -145,3 +145,26 @@ cdef SIZE_t ravel_multi_index_cython(SIZE_t[:] coords, const SIZE_t[:] shape) no
             flat_index *= shape[i + 1]
 
     return flat_index
+
+
+cdef INT32_t find_index(DTYPE_t[:] feature_values, DOUBLE_t threshold) noexcept nogil:
+    """Finds the index of the first element in `feature_values` that is greater than `threshold`.
```
adam2392: I think I had a previous comment here that was not addressed. Is this code used? If not, we don't want unused code snippets added to the codebase for no reason.

examples/plot_extra_oblique_random_forest.py
```diff
@@ -0,0 +1,170 @@
...
+
+A performance comparison between extra oblique forest and standard oblique random
+forest using four datasets from OpenML benchmarking suites.
+
+Extra oblique forest uses extra oblique trees as base model which differ from classic
+decision trees in the way they are built. When looking for the best split to
+separate the samples of a node into two groups, random splits are drawn for each
+of the `max_features` randomly selected features and the best split among those is
+chosen. When `max_features` is set 1, this amounts to building a totally random
+decision tree. For details of the algorithm, see [1]_.
```
adam2392:
```suggestion
decision trees in the way they are built. When looking for the best split to
separate the samples of a node into two groups, random thresholds are drawn for each
of the `max_features` randomly selected features and the best split among those is
chosen. This is in contrast with the greedy approach, which evaluates the best possible
threshold for each chosen split.

When `max_features` is set 1, this amounts to building a totally random
oblique decision tree. For details of the original algorithm, see [1]_.
```

examples/plot_extra_oblique_random_forest.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+params = {
+    "max_features": None,
+    "n_estimators": 50,
+    "max_depth": 5,
+    "random_state": random_state,
+    "n_cv": 10,
+    "n_repeats": 1,
+}
```
adam2392: You should add a note here regarding max_depth and the n_estimators being set low to allow the example to run on the CI. You can add some intuition on how these should be set normally.

adam2392 (changes requested):
Okay from the exps, the general code seems fine!

There are a few issues tho to be addressed in order for this to get merged:

1. unused code: if code is not used, it should not be committed. If I'm misunderstanding if certain pieces of code are actually used, lmk.
2. examples: The examples can be improved: https://output.circle-artifacts.com/output/job/80903942-02eb-4fcd-a7ca-0d25ec870888/artifacts/0/dev/auto_examples/plot_extra_oblique_random_forest.html#sphx-glr-auto-examples-plot-extra-oblique-random-forest-py. 

the examples should clearly convey some concepts. Right now, they are reworded versions of other examples. The example should guide the user to learning, or understanding something. 

The iris example should clearly convey the differences between normal oblique decision trees and extra oblique decision trees. I.e. the only difference is the randomly chosen thresholds vs picking best thresholds and the trees are still useful.

The cc-18 example should clearly convey the tradeoff between choosing extra vs normal oblique decision trees.

adam2392:
@SUKI-O I cleaned up the PR to remove any unnecessary code or documentation: https://github.com/neurodata/scikit-tree/pull/75/commits/d47f71842594b82527694188c9d1d61bb52ba296

Lmk if this looks good to you and I will merge?

adam2392:
I will merge this in for now since this PR has the biggest diff and there are other PRs on the queue, but if there are questions, or comments you have, feel free to open up a GH issue.

adam2392 (approved):

adam2392:
Thanks @SUKI-O !

PR #83, opened 2023-06-12
[ENH] MI and CMI simulation module and relevant estimators from the forest

### LINKED ISSUE #48: [PAPER] Demonstrate usage of Unsupervised tree to estimate MI and CMI
Using the unsupervised trees, I will sketch out an example estimating MI and CMI for a variety of simulation cases.

todo:

- [ ] Implement the marginalization of a forest 

"e can compute the distance between two points i, j on a subspace S of X given the forest learned on X."

- set `c` as the maximum distance between two points in the subspace S, and recompute dF(si,sj) for each pair i,j. To compute dF for a subspace S, we first need to recompute all leave assignments. That is, given a tree Tj, we assign each point i to that leaf in Tj, to which it would be assigned, if projected onto S. After reassigning the leaves, we can compute dF (si, sj ) as above. Based on the above procedure, we can compute dG(xi, xj ) and dG(yi, yj ) for each pair i,j and use these distances to compute

So once we train on:

### LINKED ISSUE #49: Add proximity matrix computation function for general trees
All trees should have access to the proximity based on the tree-distance, whether it is supervised, or unsupervised.

However, I don't want to flood the existing fork of sklearn we have as it is already a big diff. So I think we can just produce a proximity_matrix computation for trees that work for all trees. Just need to adhere to the tree API and implement the `apply` function.

We can just take what is currently implemented in `UnsupervisedDecisionTree` and refactor it out as a function and have the corresponding API in `UnsupervisedDecisionTree` call that new function.

adam2392:
Fixes #49 
Fixes #48 
Supersedes: #58 

- simulation for MI and CMI data
- 

## Before submitting

## After submitting


sktree/tree/_marginalize.py
adam2392, line 1: @YuxinB @PSSF23 @jovo in case you're interested, this is the method for marginalizing a fitted tree using the ideas that have been discussed.

This file and `_marginal.pyx` are the core files.

Specifically, I implement some ideas for traversing the tree, while "marginalizing out features".

adam2392:
Next, I need to add now a simulation using a fitted forest and computing a bunch of quantities for MI/CMI

PR #86, opened 2023-06-14
Add multi-output support to honest trees

### LINKED ISSUE #85: Add multi-output support to honest trees
Add multi-output support to `HonestTreeClassifier` & `HonestForestClassifier`

PSSF23:
Fixes #85 

- Add multi-output support to `HonestTreeClassifier` & `HonestForestClassifier`

## Before submitting

## After submitting


PSSF23 (commented):
I have removed the test skipping except `check_classifiers_multilabel_output_format_predict_proba`, which is inconsistent with how `sklearn.ensemble.RandomForestClassifier` produces `predict_proba` results. 

Error states that the shape of `(n_samples, n_outputs)` should be expected, but `(n_outputs, n_samples, n_classes)` is the actual shape produced by `sklearn` forests and trees.

Other multi-output cases passed.

adam2392:
> I have removed the test skipping except `check_classifiers_multilabel_output_format_predict_proba`, which is inconsistent with how `sklearn.ensemble.RandomForestClassifier` produces `predict_proba` results.
> Error states that the shape of `(n_samples, n_outputs)` should be expected, but `(n_outputs, n_samples, n_classes)` is the actual shape produced by `sklearn` forests and trees.

`(n_samples, n_outputs)` is the shape we want/required, but `(n_outputs, n_samples, n_classes)`  is what we get for the honest tree/forest?

I'll take a look.

In the meantime, can you see if you can think of a quick test we can do to judge the multi-output validity? I.e. Perhaps test that the forest gets the correct answer on a short simulation? You can use the existing tests as inspiration.

I wonder if we can extend the existing figure that's shown in the uncertainty forest paper to have a multi-output setting.

PSSF23:
I don't think `(n_samples, n_outputs)` is a valid output format for `predict_proba`, and that's not the output of current sklearn implementations. It seems to be more appropriate for results of `predict`, where we get class labels instead of posteriors. Maybe the check is outdated. I'll look into possible tests for multi.

adam2392:
https://github.com/neurodata/scikit-tree/pull/86/commits/f1ab592c9d01e0e8a6f5a514f69d5413754b98fc fixes the issue with the `check_estimator` test with multi-output predict_proba. 

It would be great if we have a simulation to verify the results of a multi-output honest classification

PSSF23:
Would that be possible as we don't have a regressor class yet?

adam2392:
> Would that be possible as we don't have a regressor class yet?

Sorry classification

sktree/ensemble/_honest_forest.py
```diff
@@ -451,14 +453,23 @@ def _predict_proba(self, X, indices=None, impute_missing=None):
         )
 
         # Normalize to unit length, due to prior weighting
-        zero_mask = posteriors.sum(1) == 0
+        posteriors = np.array(posteriors)
+        zero_mask = posteriors.sum(2) == 0
         posteriors[~zero_mask] /= posteriors[~zero_mask].sum(1, keepdims=True)
+
         if impute_missing is None:
             posteriors[zero_mask] = self.empirical_prior_
```
PSSF23: Having trouble on this line. It seems that when defined, `empirical_prior_` doesn't have dimensional differences for `y` labels. So I'll try to increase its dimensions.

PSSF23: Actually, why is the empirical prior set at the forest level. It's already set in honest trees, so this part can be removed.

sktree/tests/test_honest_forest.py
```diff
@@ -59,6 +59,37 @@ def test_iris(criterion, max_features, estimator):
...
+        n_estimators=10,
+        tree_estimator=estimator,
+    )
+
+    second_y = np.concatenate([(np.ones(10) * 3), (np.ones(20) * 4), (np.ones(120) * 5)])
+
+    X = iris.data
+    y = np.stack((iris.target, second_y)).T
+    clf.fit(X, y)
+    score = mean_squared_error(clf.predict(X), y)
```
adam2392: Why not just use https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html

and assert some non-trivial performance? 

Is it possible to interpret MSE of 0.5-1.0?

PSSF23: It appears that `accuracy_score` doesn't support multi-output. It seems that the default way they measure such predictions is through `sklearn.multioutput.MultiOutputClassifier`.

PSSF23:
https://github.com/neurodata/scikit-learn/pull/45#pullrequestreview-1483593227

PSSF23 (commented):
Update the test to `r2_score` as that's the default `sklearn` uses for multi-output scores. Also included different `honest_prior` options.

sktree/tests/test_honest_forest.py
```diff
@@ -84,10 +88,15 @@ def test_iris_multi(criterion, max_features, estimator):
...
-    )
+    score = r2_score(clf.predict(X), y)
+    if honest_prior == "ignore":
+        assert score < 1.0, "Failed with {0}, criterion = {1} and score = {2}".format(
+            "HForest", criterion, score
+        )
+    else:
+        assert (
+            score > 0.0 and score < 1.0
+        ), "Failed with {0}, criterion = {1} and score = {2}".format("HForest", criterion, score)
```
adam2392: Does this actually test anything? Isn't r2_score between 0 and 1 always?

PSSF23: No it's negative if the model performs worse than chance.

adam2392: Can we make the test stronger? Or does it fail? E.g. can we increase the lower bound by some amount?

```
a constant model that always predicts the average y disregarding the input features would get a 
R^2 score of 0.0.
```

So ideally, we have some guarantee of performance on e.g. the iris multi-output problem. Otw, we could accidentally screw something up and not catch it.

PSSF23: Sure, we can have a tighter score. The current second dimension of y is not very informative, and that's probably why the score is low. I will try to optimize that.

PSSF23 (commented):
I fixed some bugs with leaf updates and the scores are much higher now.

PSSF23:
It seems that `_impute_missing_classes` is really not used... Should we just remove it?

PSSF23:
@adam2392 can you make a commit to include the submodule changes in here? Just want to check if the problem is fixed.

adam2392:
> @adam2392 can you make a commit to include the submodule changes in here? Just want to check if the problem is fixed.

Run 

```
git submodule update --recursive --remote
git add -A
git commit -m "Update submodule commit ID"

and then

./spin build -j 4 --forcesubmodule
```

This updates the  submodule and then rebuilds the package locally.

adam2392:
> It seems that `_impute_missing_classes` is really not used... Should we just remove it?

This is because the honest trees always get all classes? Can we up the honest_ratio and this will get activated?

Otw... sure you can delete if you think that's fine. I just want to make sure this actually works in relevant edge cases.

adam2392:
Looks like some other issues to fix with the docstrings in sklearn fork. I can do that later unless you do a quick fix now

adam2392:
The circleCI docs build should be fixed now @PSSF23

PSSF23 (commented):
@adam2392 you reversed one of the changes in the honest tree for multi-output, which I updated back in the last commit.

adam2392:
> @adam2392 you reversed one of the changes in the honest tree for multi-output, which I updated back in the last commit.

Oh apologies! Messed up the merge commit most likely

sktree/tree/_honest_tree.py
```diff
@@ -431,29 +425,31 @@ def _inherit_estimator_attributes(self):
         self.n_outputs_ = self.estimator_.n_outputs_
         self.tree_ = self.estimator_.tree_
 
-    def _empty_leaf_correction(self, proba, normalizer):
+    def _empty_leaf_correction(self, proba, pos=0):
```
adam2392, line 428: Can you just add a short docstring to describe what's going on? I'm reading these lines and having trouble figuring out what `pos` does actually :/

sktree/tests/test_honest_forest.py
```diff
@@ -59,7 +64,49 @@ def test_iris(criterion, max_features, estimator):
...
+            score > 0.6 and score < 1.0
+        ), "Failed with {0}, criterion = {1} and score = {2}".format("HForest", criterion, score)
+    else:
+        assert (
+            score > 0.9 and score < 1.0
+        ), "Failed with {0}, criterion = {1} and score = {2}".format("HForest", criterion, score)
+
+
 def test_impute_classes():
+    # This test is unnecessary as sample_weight method doesn't impute
```
adam2392: Do we want to keep or remove this test?

adam2392 (approved):
I had two questions that I left.

Otherwise, LGTM to merge once those are addressed!

sktree/tree/_honest_tree.py
PSSF23: `pos` indicates the class dimension of y, so that posterior corrections only work on that dimension in multi-output cases.

sktree/tree/_honest_tree.py
```diff
@@ -431,29 +425,36 @@ def _inherit_estimator_attributes(self):
...
         self.tree_ = self.estimator_.tree_
 
-    def _empty_leaf_correction(self, proba, normalizer):
-        """Leaves with empty posteriors are assigned values"""
+    def _empty_leaf_correction(self, proba, pos=0):
+        """Leaves with empty posteriors are assigned values.
+
+        The posteriors are corrected according to the honest prior.
+        In multi-output cases, the posterior corrections only correspond
+        to the respective y dimention, indicated by the position param pos.
```
adam2392:
```suggestion
        to the respective y dimension, indicated by the position param pos.
```

PSSF23:
This `test_max_samples` test is very weird, succeeding or failing randomly.

adam2392:
> This `test_max_samples` test is very weird, succeeding or failing randomly.

There is no random_state parameter passed to the `HonestForestClassifier` and that is perhaps what's causing this? Can we add that in since that will be definitely required?

adam2392:
If it is failing, does that mean the tree depths is some edge case?

PSSF23:
> If it is failing, does that mean the tree depths is some edge case?

Maybe just the samples were not enough for the tree to split

adam2392:
It looks like there is a depth of -1. How does that even occur?...

```
FAILED sktree/tests/test_honest_forest.py::test_max_samples - assert False
 +  where False = all(array([ 7, -1]) > 0)
 +    where array([ 7, -1]) = <function diff at 0x1102396b0>([1, 8, 7])
 +      where <function diff at 0x1102396b0> = np.diff
```

PSSF23:
> It looks like there is a depth of -1. How does that even occur?...
> 
> ```
> FAILED sktree/tests/test_honest_forest.py::test_max_samples - assert False
>  +  where False = all(array([ 7, -1]) > 0)
>  +    where array([ 7, -1]) = <function diff at 0x1102396b0>([1, 8, 7])
>  +      where <function diff at 0x1102396b0> = np.diff
> ```

That occurs when the tree did not split. root has depth -1.

PSSF23 (commented):
I added the random seed for X generation as well, so the test results should be consistent. Overall, when the sample size is too small, some honest trees just wouldn't work as intended.

adam2392:
I see since the root does not split at all, than the max_depth is -1, rather than 1, 2, ... and so on.

sktree/tests/test_honest_forest.py
```diff
@@ -108,6 +108,7 @@ def test_iris_multi(criterion, max_features, honest_prior, estimator):
 def test_max_samples():
     max_samples_list = [8, 0.5, None]
     depths = []
+    np.random.seed(0)
```
adam2392: Don't set global random seed. I just realized a lot of tests are doing this. This actually isn't thread safe which is an issue for Cythonized code.

Instead you can set a global `seed = 12345` and then for each place-in for `np.random.`, you run `rng = np.random.default_rng(seed)` and use `rng` in place of `np.random`

Can you do this everywhere in the file (I think just 8 places that uses `np.random.seed`)? 

Here's a ref talking about it: https://albertcthomas.github.io/good-practices-random-number-generators/

PSSF23: will do.

adam2392:
One minor issue I caught just now, but otw then this LGTM

adam2392 (approved):

adam2392:
Thanks @PSSF23 !

PR #89, opened 2023-06-22
modified:   sktree/tree/manifold/_morf_splitter.pyx

YuxinB:
Modify the morf_splitter for multi-view of MORF.

### DIFF
--- a/sktree/tree/manifold/_morf_splitter.pyx
+++ b/sktree/tree/manifold/_morf_splitter.pyx
@@ -27,6 +27,7 @@ cdef class PatchSplitter(BaseObliqueSplitter):
     def __cinit__(
         self,
         Criterion criterion,
+        int n_features_1,
         SIZE_t max_features,
         SIZE_t min_samples_leaf,
         double min_weight_leaf,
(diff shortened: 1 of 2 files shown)


YuxinB:
Added one new attributes named n_features_1, and change  sample_proj_mat code.  my problem is, If the change below is doable, how to define n_features_1 in classes, can I set it as an input?

sktree/tree/manifold/_morf_splitter.pyx
```diff
@@ -281,7 +283,17 @@ cdef class BestPatchSplitter(BaseDensePatchSplitter):
         # (\prod_i self.patch_dims_buff[i])
         cdef SIZE_t patch_size
 
-        for proj_i in range(0, max_features):
+
+        ##### suposse randomly select one feature set to find a best split
+        cdef np.ndarray[int, ndim=1] choices = np.array([0, n_features_1])
```
adam2392, line 288: This doesn't work inside `nogil`. To instantiate an array requires the GIL. 

I thought what you are trying to do is pass in `n_features_1 = the # of features in dataset 1`. Then `X[:, :n_features_1]` is dataset 1 and `X[:, n_features_1:]` is dataset 2?

YuxinB: yes! Sorry, I didn't check the git timely

PR #125, opened 2023-09-11
[ENH] add PermutationForests and FeatureImportanceForests to sktree

### LINKED ISSUE #111: Posterior Forests (or whatever we call them)
I'd like the docs to include very explicitly and at a high-level, that we can estimate posteriors, entropy, conditional entropy, MI, and eventually CMI

### LINKED ISSUE #112: add multiview code/docs/tutorial

### LINKED ISSUE #120: pAUC & pvalue
i also want docs/tutorials showing how we compute pAUC
and another one showing how we compute a p-value

PSSF23:

- Related to #111 

#112 , #120 to be addressed in a future PR

## Before submitting

## After submitting


sampan501:
Will this also have MIRF with Mutual Info as a stat?

PSSF23 (commented):
Referring to [this file](https://github.com/neurodata/hyppo/blob/1642f769356a6a064bb4d579fb25d1c22fd0ac1b/hyppo/independence/mirf.py) in `hyppo`. I removed everything related to `hyppo` for now, including the mutual information function, to avoid confusion. The `_might.py` file includes all other single-feature and multi-view methods we developed.

adam2392:
> Referring to [this file](https://github.com/neurodata/hyppo/blob/1642f769356a6a064bb4d579fb25d1c22fd0ac1b/hyppo/independence/mirf.py) in `hyppo`. I removed everything related to `hyppo` for now, including the mutual information function, to avoid confusion. The `_might.py` file includes all other single-feature and multi-view methods we developed.

For now I will remove the multi-view stuff prolly and then also look at how we can introduce arbitrary metrics in here: e.g. MI, ROC_auc, etc.

I'll take a look tonight

PSSF23 (commented):
Per @sampan501 , added `stat="MI"` and `stat="AUC"` as parameters for `statistic()`.

sampan501:
@PSSF23 Can you also rename it to MIGHT? Might make all this name changing easier to follow - no pun intended

PSSF23 (commented):
- Renamed to `MIGHT` and `MIGHT_MV`
- Added y-label permutation test to `MIGHT` (previously removed due to connection to `hyppo`)

PSSF23:
@sampan501 I might be misunderstanding the MI calculation, but why the original method had the `axis=1` param? If it's not needed I'll correct `MIGHT_MV` as well.

sampan501:
`predict_proba` returns a `(n_samples, n_classes)` array as an output. So, the previous MI calculation was taking averages over the classes. We don't need to do that since you do that already in `forest_pos`. We can probably remove the mean for both calculations too.

sampan501:
Before we make any serious changes like the one above, we really need unit tests for this method

PSSF23 (commented):
I made an empty test file. Let's edit and add to this message about what we need to test:

- [ ] test on iris for mutual info for accuracy
- [ ] partial AUROC
- [ ] multiview splitter (not this time)

adam2392:
> I made an empty test file. Let's edit and add to this message about what we need to test:
> 
> * [ ]  test on iris for mutual info for accuracy
> * [ ]  partial AUROC
> * [ ]  multiview splitter

Let's do multi view in a sep PR.

PSSF23 (commented):
By unit test results, MIGHT seems to perform the worst when coupled with `PatchObliqueDecisionTreeClassifier()`. I remember similar situations back with honest tree tests. Should I lower the passing threshold or remove the estimator option?

sktree/stats/_might.py
```diff
@@ -0,0 +1,317 @@
...
+            H_Y = entropy(counts, base=np.exp(1))
+            self.stat = max(H_Y - H_YX, 0)
+
+        if return_pos:
+            return self.stat, posterior
+
+        return self.stat
+
+    def test(self, x, z, y, reps=1000, workers=1, random_state=None):
+        XZ = np.hstack((x, z))
```
adam2392: I realize there's no "right way" to do this, but I went for the strategy of passing in one data array and then index array. This removes the necessity of creating a concatenation of two potentially very large arrays, which i) increases RAM usage and ii) can add to the computation time.

sktree/stats/_might.py
```diff
@@ -0,0 +1,317 @@
...
+        return_pos=False,
+        permute_y=False,
+    ):
+        # Initialize trees
+        if initial:
+            self.clf.fit(x[0:2], y.ravel()[0:2])
+
+        # Compute posteriors with train test splits
+        posterior = Parallel(n_jobs=workers)(
+            delayed(auc_calibrator)(tree, x, y.ravel(), test_size, permute_y)
```
adam2392: I'm not sure I understand why `y` would ever get permuted

sktree/stats/_might.py
```diff
@@ -0,0 +1,317 @@
...
+
+def forest_pos(posterior, y):
+    # Average all posteriors
+    posterior_final = np.nanmean(posterior, axis=0)
+
+    # Ignore all NaN values (samples not tested)
+    true_final = y.ravel()[~np.isnan(posterior_final)].reshape(-1, 1)
+    posterior_final = posterior_final[~np.isnan(posterior_final)].reshape(-1, 1)
+
+    return np.hstack((true_final, posterior_final))
```
adam2392: This does not work and also would introduce a bug if `y` is `(n_samples, n_outputs > 1)`?

adam2392:
> Changes proposed in this pull request:
> 
> * Fix [Posterior Forests (or whatever we call them) #111](https://github.com/neurodata/scikit-tree/issues/111)
> * Fix [add multiview code/docs/tutorial #112](https://github.com/neurodata/scikit-tree/issues/112)
> * Fix [pAUC & pvalue #120](https://github.com/neurodata/scikit-tree/issues/120)

I would remove the second two bullets to prevent those issues from getting closed w/o actually resolving the issue raised.

adam2392 (commented):
Left a few comments that I'm trying to understand. I put up two implementation files:

- `forestht.py`
- `utils.py`

The main class is called `ForestHT`, which encompasses both the Coleman and permutation approach to hypothesis testing w/ the forest as I understand it.

It allows generally any `metric` that follows scikit-learn API of `def metric(y_true, y_pred, ...)`. I'm in favor of this if it LGT you.

sktree/stats/_might.py
PSSF23: This is another design that jovo came up, which has unique y-label permutation for individual trees.

sktree/stats/_might.py
PSSF23: You are right. This is only limited to 2 class problems right now.

PSSF23:
> > Changes proposed in this pull request:
> > 
> > * Fix [Posterior Forests (or whatever we call them) #111](https://github.com/neurodata/scikit-tree/issues/111)
> > * Fix [add multiview code/docs/tutorial #112](https://github.com/neurodata/scikit-tree/issues/112)
> > * Fix [pAUC & pvalue #120](https://github.com/neurodata/scikit-tree/issues/120)
> 
> I would remove the second two bullets to prevent those issues from getting closed w/o actually resolving the issue raised.

Feel free to do so. I overestimated the scope of this PR.

sktree/stats/_might.py
adam2392: But this then does not work for the problem that the Coleman paper addresses right?

If we are interested in whether or not X1 is important for predicting Y in terms of AUC, for X = [X1, X2, ..., Xd], then permuting Y destroys the dependency of Y on all other X?

Or perhaps this design was to address the problem of whether or not generally X (the whole column space) provides information on Y? If so, then sure this is valid, but if we permute the row indices of X, then it can cover both use-cases.

cc: @jovo

PSSF23: I assume this is only for null observations, but yeah on multi-view it would not make sense.

sktree/stats/forestht.py
```diff
@@ -0,0 +1,466 @@
...
+    def _statistic(
+        self,
+        estimator: ForestClassifier,
+        X: ArrayLike,
+        y: ArrayLike,
+        covariate_index: ArrayLike = None,
+        metric="auc",
+        test_size=0.2,
+        **metric_kwargs,
+    ):
```
PSSF23: I think it will still be a good optional functionality to output the posteriors, so we can save time on potential metric modifications in the future.

sktree/stats/_might.py
jovo: the key is for each tree to do the permutation independently, Coleman only did 1 permutation.  i do not feel strongly about whether to permute X or Y, for our test, it doesn't matter.

sktree/stats/forestht.py
adam2392: Added

adam2392:
The current implementation "sort of" seems to work for the "correlated logit" model from the Coleman paper (see unit-test). However, when we test it with AUC and iris, the results are not clear. There could be an issue w/ the implementation cc: @PSSF23 and @sampan501 

However some things we can try are:

- reproduce the other simulations in Coleman et al., which require us to use a ForestRegressor (i.e. not the HonestForestClassifier object), since those simulations have `y` being continuous.
- separate training/testing set for the forest instead of at the tree levels independently. Right now, the posteriors are weirdly very extreme (e.g. either 0 or 1) in the iris example.

Feel free to experiment and I'll revisit this later on.

sktree/stats/utils.py
```diff
@@ -0,0 +1,243 @@
...
+    X = np.concatenate((X_train, X_test), axis=0)
+    index_arr = np.arange(X.shape[0], dtype=int)  # .reshape(-1, 1)
+
+    null_metrics = np.zeros((n_repeats,))
+
+    for idx in range(n_repeats):
+        # permute the covariates inplace
+        rng.shuffle(test_index_arr)
+        perm_X_cov = X_test[test_index_arr, covariate_index]
+        X_test[:, covariate_index] = perm_X_cov
```
PSSF23, line 152: Why is the test set shuffled?

sktree/stats/utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+from sktree._lib.sklearn.ensemble._forest import ForestClassifier
+from sktree._lib.sklearn.tree import DecisionTreeClassifier
+
+
+def _mutual_information(y_true, y_pred):
+    H_YX = np.mean(entropy(y_pred, base=np.exp(1)))
+    _, counts = np.unique(y_true, return_counts=True)
+    H_Y = entropy(counts, base=np.exp(1))
+    return max(H_Y - H_YX, 0)
```
PSSF23: This is wrong. The method in `_might.py` is buggy. The correct one is in `hyppo`.

sktree/stats/permutationforest.py
```diff
@@ -0,0 +1,835 @@
...
+        indices = np.arange(n_samples, dtype=int)
+
+        if covariate_index is not None:
+            # perform permutation of covariates
+            index_arr = rng.choice(indices, size=(X.shape[0], 1), replace=False, shuffle=False)
+            X = X.copy()
+            X[:, covariate_index] = X[index_arr, covariate_index]
+
+        X_train, X_test = X[self.indices_train_, :], X[self.indices_test_, :]
+        y_train, y_test = y[self.indices_train_, :], y[self.indices_test_, :]
```
PSSF23: So this method only permutes the null individually but maintains the common training and test sets?

sktree/stats/utils.py
adam2392: I'll replace. I don't use this method currently.

sktree/stats/permutationforest.py
adam2392: Yeah I think that is the way it was done in the paper, but we can try it out w/ permuting everything as you had in `_might.py`, but just figured this should at least reproduce what was in Coleman, albeit not as "good results" it seems. 

Their simulations are a bit hard to read and interpret too

sktree/stats/utils.py
adam2392: The only thing is `axis=1` within the entropy computation for H_YX right?

PSSF23: As you already implemented the multi-class, this should fix it.

doc/whats_new/v0.2.rst
```diff
@@ -32,6 +32,7 @@ Changelog
 - |API| Allow ``sqrt`` and ``log2`` keywords to be used for ``min_samples_split`` parameter in unsupervised forests, by `Adam Li`_  (:pr:`114`)
 - |Feature| Implement extended isolation forest, by `Adam Li`_ (:pr:`101`)
 - |Feature| Implementation of StreamDecisionForest, by `Haoyin Xu`_ and `Adam Li`_ (:pr:`116`)
+- |Feature| Implementation of Permutation forests and a feature importance testing forest, by `Haoyin Xu`_, `Adam Li`_, `Sambit Panda` (:pr:`125`)
```
adam2392, line 35: @PSSF23 and @sampan501 feel free to directly add others who added to the Hyppo tree implementation

sktree/stats/_might.py
adam2392, line 1: @PSSF23 once you review the code and we get MI parity, we can perhaps remove this file?

Besides computing the MI/P-AUC test statistic, am I missing anything else?

doc/whats_new/v0.2.rst
sampan501: I think that's everyone

sktree/stats/_might.py
PSSF23: Yeah I think everything else is addressed.

adam2392 (approved):
Once CI passes, this is my final push. Any other steps are to extend how we can utilize the MI, or examples.

examples/plot_MI_gigantic_hypothesis_testing_forest.py
```diff
@@ -0,0 +1,133 @@
...
+    sample_dataset_per_tree=True,
+)
+
+# we test for the first feature set, which is important and thus should return a pvalue < 0.05
+stat, pvalue = est.test(
+    X, y, covariate_index=np.arange(n_features_set, dtype=int), metric="mi", n_repeats=n_repeats
+)
+print(f"Estimated MI difference: {stat} with Pvalue: {pvalue}")
+
+# we test for the second feature set, which is important and thus should return a pvalue > 0.05
```
adam2392:
```suggestion
# we test for the second feature set, which is unimportant and thus should return a pvalue > 0.05
```

sktree/stats/forestht.py
adam2392, line 1: We should specify in the docstring what results are "cached". Since the OG forest is cached (X, y), then we don't need to recompute the non-permuted test statistic. However, the permuted forest is recomputed if `covariate_index` changes of course.

We should add checks to make sure i) the data is the same in terms of `n_samples`, and `n_features`, else it is a different `X` altogether. We also want to perhaps use `type_of_target ` to check that `y` is the same type of target as well. We could cache `covariate_index` as `_covariate_index_cache` to make it easy to re-grab the pvalue by not training the permuted forest again.

We should add some unit-tests to check all this behavior too.

benchmarks_nonasv/test_permutation_forest.ipynb
adam2392, line 1: This file can be removed, as it's just a notebook experimenting w/ producing figures relevant for reproducing Coleman et al.

sampan501 (approved):

sktree/stats/forestht.py
```diff
@@ -0,0 +1,736 @@
...
+            estimator_ = self.estimator
+        return estimator_
+
+    def _statistic(
+        self,
+        estimator: ForestClassifier,
+        X: ArrayLike,
+        y: ArrayLike,
+        covariate_index: ArrayLike = None,
+        metric="mse",
```
PSSF23: We should set the default metrics to AUC or MI, as current paper doesn't use MSE.

adam2392: Done

benchmarks_nonasv/test_permutation_forest.ipynb
adam2392: Actually I put this into a sep folder to keep around for note-keeping

sktree/stats/forestht.py
adam2392: This was addressed for `FeatureImportance*Forest`

adam2392: Caching covariate index actually didn't work that nicely, so I think it's not a big deal to not have that feature. + the fitted statistic/null is saved, so the pvalue is easily recomputed without actually calling `test()`.

adam2392:
@YuxinB just a gentle ping to see if you have the time to do a code-review

PSSF23 (commented):
@adam2392 Correct me if I'm wrong, right now `self.posterior_final_` only saves the forest posteriors? But the Coleman method requires tree posteriors.

sktree/stats/utils.py
```diff
@@ -0,0 +1,246 @@
...
+    Parameters
+    ----------
+    X_test : ArrayLike of shape (n_samples, n_features)
+        The data matrix.
+    y_test : ArrayLike of shape (n_samples, n_outputs)
+        The output matrix.
+    y_pred_proba_normal : ArrayLike of shape (n_samples_normal, n_outputs)
+        The predicted posteriors from the normal forest.
+    y_pred_proba_perm : ArrayLike of shape (n_samples_perm, n_outputs)
+        The predicted posteriors from the permuted forest.
```
PSSF23: These should be of shape `(n_trees, n_samples, n_outputs)` @adam2392

PSSF23: When `sample_dataset_per_tree` is on, both arrays could have NaN values.

YuxinB: Will review this afternoon, sorry about the delay

sktree/stats/forestht.py
```diff
@@ -0,0 +1,794 @@
...
+        return_posteriors: bool = False,
+        **metric_kwargs,
+    ):
+        """Helper function to compute the test statistic."""
+        metric_func = METRIC_FUNCTIONS[metric]
+        rng = np.random.default_rng(self.random_state)
+
+        if self.permute_per_tree:
```
PSSF23: This looks to me should be `if self.sample_dataset_per_tree`.

sktree/stats/forestht.py
```diff
@@ -0,0 +1,794 @@
...
+        """Helper function to compute the test statistic."""
+        metric_func = METRIC_FUNCTIONS[metric]
+        rng = np.random.default_rng(self.random_state)
+
+        if metric in POSTERIOR_FUNCTIONS:
+            predict_posteriors = True
+        else:
+            predict_posteriors = False
+
+        if self.permute_per_tree:
```
PSSF23: Same as comment above, should be `if self.sample_dataset_per_tree`

sktree/stats/forestht.py
```diff
@@ -0,0 +1,794 @@
...
+                metric=metric,
+                n_repeats=n_repeats,
+                seed=self.random_state,
+            )
+        else:
+            if not self.sample_dataset_per_tree:
+                _, indices_test = self.train_test_samples_[0]
+                y_test = y[indices_test, :]
+            else:
+                y_test = y
```
PSSF23: The first if statement should be `if not self.permute_per_tree`? The second condition now is redundant.

PSSF23:
It seems that when I use `FeatureImportanceForestClassifier` it will always default to `RandomForestClassifier` even when I set `estimator=HonestForestClassifier`. @adam2392

adam2392:
> It seems that when I use `FeatureImportanceForestClassifier` it will always default to `RandomForestClassifier` even when I set `estimator=HonestForestClassifier`. @adam2392

Do you have some code for this?

PSSF23:
The performance between `FeatureImportanceForestClassifier` and any `HonestForestClassifier` methods are just too different.

This is 500 samples (50 training), MI / Tree is FI, rest are honest trees.
overlapping_gaussians_4096_500

This is 6000 samples (600 training) using FI, and the bias starts to show. It's similar to a bug I saw previously, when the predicted class labels are used instead of true posteriors. There are many zeros and ones in the posterior, so I believe they are generated by `DecisionTreeClassifier`.
overlapping_gaussians_4096_6000

adam2392:
Do you have the code you used for this? Can you send to me? @PSSF23

PSSF23 (commented):
The error message:
```python
FAILED sktree/stats/tests/test_forestht.py::test_featureimportance_forest_permute_pertree - ValueError: all the input array dimensions except for the concatenation axis must match exactly, but along dimension 2, the array at index 0 has size 2 and the array at index 1 has size 1
```

adam2392:
Wanna give this one last glance @PSSF23 @sampan501 @YuxinB ?

sktree/stats/forestht.py
```diff
@@ -0,0 +1,862 @@
...
+        if predict_posteriors:
+            # now initialize posterior array as (n_trees, n_samples_test, n_classes)
+            posterior_arr = np.full(
+                (self.n_estimators, self._n_samples_, estimator.n_classes_), np.nan
+            )
+        else:
+            # now initialize posterior array as (n_trees, n_samples_test, n_outputs)
+            posterior_arr = np.full(
+                (self.n_estimators, self._n_samples_, estimator.n_outputs_), np.nan
+            )
```
PSSF23: @adam2392 I don't see why the statistic metric would affect the shape of posteriors, and the first case does not cover multi-output situations. The shape should be like: 
```python
(self.n_estimators, self._n_samples_, estimator.n_outputs_, estimator.n_classes_[0])
```

PSSF23 (commented):
In FI, I cleared all the duplicate variables like `posteriors_final_` and `observe_posteriors_`, and save the permuted statistic as class variable `permute_stat_` (all other results of permutation are saved, so this one should be there as well). I also make all default statistic MI, but we might change it to pAUC later.

sktree/stats/forestht.py
adam2392, line 1: This will require us to override the `def statistic` for Forest*Regressor as `mi` is not supported there.

sktree/stats/forestht.py
adam2392: If we don't predict the posteriors, we simply predict the class (i.e. via predict).

If we predict the posteriors, you're right this does not support multi-output situations right now, but I presumed that was fine? 

> (self.n_estimators, self._n_samples_, estimator.n_outputs_, estimator.n_classes_[0])

Would not be sufficient. Each output may have a different number of classes, so the strategy in scikit-learn is to then return a list of arrays instead, since the last dimension can vary.

sktree/stats/forestht.py
adam2392: Resolved.

sktree/stats/forestht.py
adam2392: Resolved.

sktree/stats/forestht.py
adam2392: Resolved.

sktree/stats/utils.py
adam2392: Resolved. I believe(?)

sktree/stats/forestht.py
adam2392: Done.

sktree/stats/tests/test_forestht.py
```diff
@@ -0,0 +1,414 @@
...
+            ),
+            test_size=0.5,
+        )
+        pvalue = clf.test(X, y, covariate_index=[covariate_index], metric="mi")
+        return pvalue
+
+    out = Parallel(n_jobs=1, backend=backend)(
+        delayed(run_forest)(covariate_index) for covariate_index in range(n_features)
+    )
+    assert len(out) == n_features
```
adam2392: @sampan501 to my knowledge, any issues w/ joblib should be fixed or are the result of some other issues perhaps? 

Lmk if this unit-test sufficiently captures what the usage looks like in the power simulations.

sampan501: `n_jobs = 1` should be `n_jobs = n_jobs` in line 411, but otherwise good

PSSF23:
@adam2392 can the print statements in FIClf be removed?

adam2392:
Yeah feel free to push that. Im making some other changes tho so anything larger feel free to open a PR to this PR.

sktree/stats/tests/test_forestht.py
```diff
@@ -205,12 +209,12 @@ def test_linear_model(hypotester, model_kwargs, n_samples, n_repeats, test_size)
                     n_jobs=-1,
                 ),
                 "random_state": seed,
-                "permute_per_tree": True,
-                "sample_dataset_per_tree": True,
+                "permute_per_tree": False,
```
sampan501: This is testing MI Sep and not MI/Tree right? Where are we testing MI/Tree?

PSSF23: This is MI / tree

PSSF23: Agreed. This is MI Sep. We should test MI / Tree as well.

adam2392: Yeah I will add a pytest.mark.parametrize in a bit. I think last night the pvalue behavior was not converging as well as the MI Sep.

sampan501:
@adam2392 @PSSF23 I'm still having an error with the sample size when using the parameters from our meeting today (n = 32, test_size = 0.2). Here's the trace:
```
"""
Traceback (most recent call last):
  File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/externals/loky/process_executor.py", line 463, in _process_worker
    r = call_item()
        ^^^^^^^^^^^
  File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/externals/loky/process_executor.py", line 291, in __call__
    return self.fn(*self.args, **self.kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/parallel.py", line 589, in __call__
    return [func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/parallel.py", line 589, in <listcomp>
    return [func(*args, **kwargs)
            ^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/mendseqs/high-d/high-d-sims.py", line 1342, in compute_null
    pval = _nonperm_pval(test, sim, n, p, noise=noise, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/mendseqs/high-d/high-d-sims.py", line 1277, in _nonperm_pval
    pvalue = test[0](**test[1]).test(u, v, **kwargs)[1]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/scikit-tree/sktree/stats/forestht.py", line 412, in test
    metric_star, metric_star_pi = _compute_null_distribution_coleman(
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/scikit-tree/sktree/stats/utils.py", line 291, in _compute_null_distribution_coleman
    first_half_metric = metric_func(y_test[non_nan_samples, :], y_pred_first_half)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/sambit/scikit-tree/sktree/stats/utils.py", line 34, in _mutual_information
    raise ValueError(f"y_true must be 1d, not {y_true.shape}")
ValueError: y_true must be 1d, not (1, 1)
"""
```

adam2392:
> @adam2392 @PSSF23 I'm still having an error with the sample size when using the parameters from our meeting today (n = 32, test_size = 0.2). Here's the trace:
> 
> ```
> """
> Traceback (most recent call last):
>   File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/externals/loky/process_executor.py", line 463, in _process_worker
>     r = call_item()
>         ^^^^^^^^^^^
>   File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/externals/loky/process_executor.py", line 291, in __call__
>     return self.fn(*self.args, **self.kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/parallel.py", line 589, in __call__
>     return [func(*args, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/miniconda3/envs/cancer/lib/python3.11/site-packages/joblib/parallel.py", line 589, in <listcomp>
>     return [func(*args, **kwargs)
>             ^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/mendseqs/high-d/high-d-sims.py", line 1342, in compute_null
>     pval = _nonperm_pval(test, sim, n, p, noise=noise, **kwargs)
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/mendseqs/high-d/high-d-sims.py", line 1277, in _nonperm_pval
>     pvalue = test[0](**test[1]).test(u, v, **kwargs)[1]
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/scikit-tree/sktree/stats/forestht.py", line 412, in test
>     metric_star, metric_star_pi = _compute_null_distribution_coleman(
>                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/scikit-tree/sktree/stats/utils.py", line 291, in _compute_null_distribution_coleman
>     first_half_metric = metric_func(y_test[non_nan_samples, :], y_pred_first_half)
>                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/data/sambit/scikit-tree/sktree/stats/utils.py", line 34, in _mutual_information
>     raise ValueError(f"y_true must be 1d, not {y_true.shape}")
> ValueError: y_true must be 1d, not (1, 1)
> """
> ```

Any chance you can reproduce the error w/ a small code snippet?


The following code works for me:

```
def test_small_dataset():
    n_samples = 32
    n_features = 5
    X = rng.uniform(size=(n_samples, n_features))
    y = rng.integers(0, 2, size=n_samples)  # Binary classification

    clf = FeatureImportanceForestClassifier(
        estimator=HonestForestClassifier(
            n_estimators=10, random_state=seed, n_jobs=1, honest_fraction=0.5
        ),
        test_size=0.2,
        permute_per_tree=False,
        sample_dataset_per_tree=False,
    )
    stat, pvalue = clf.test(X, y, covariate_index=[1,2], metric='mi')
```

adam2392:
FYI, I added a short unit-test to test small sample-sizes.

sampan501:
I will once I find the simulation and sample size that's causing the issue

PR #133, opened 2023-10-04
Remove numpy nanmean warnings and also bug fix of some code

adam2392:
Fixes #

-

## Before submitting

## After submitting


PSSF23 (commented):
Is there reason why the auc metric has the tree posterior shape change? Having a binary check is fine, but `roc_auc_score` can handle two classes.

adam2392:
> Is there reason why the auc metric has the tree posterior shape change? Having a binary check is fine, but `roc_auc_score` can handle two classes.

No. The reason is just to limit functionality while we ensure this works. We can remove those with a minor refactor easily. Rn the code has an inline comment reminding us that this only works for n_outputs==1.

PR #143, opened 2023-10-12
Stratify sampling when split train/test data

YuxinB:
Fixes #

-

## Before submitting

## After submitting


sktree/stats/forestht.py
```diff
@@ -206,13 +206,16 @@ def _get_estimators_indices(self, sample_separate=False):
...
+
             )
+            
             for _ in self.estimator_.estimators_:
                 yield indices_train, indices_test
 
-    @property
-    def train_test_samples_(self):
+    
+    def train_test_samples_(self,y):
```
adam2392: If y will be passed. There should be a test that y is of the same type during fitting. 

I.e. type of target, n samples, n outputs should be invariant.

adam2392 (changes requested):
If public API changes we need a test to control accidental misuse

PSSF23 (changes requested):
The changes I marked are likely regressions and should be removed.

sktree/stats/forestht.py
```diff
@@ -463,9 +468,6 @@ def test(
             observe_stat = self.observe_stat_
 
         # next permute the data
-        if covariate_index is None:
-            covariate_index = np.arange(X.shape[1], dtype=int)
-
```
PSSF23, line 468: These changes are regressions and should be removed

sktree/stats/forestht.py
```diff
@@ -332,6 +335,9 @@ def statistic(
         if self._type_of_target_ is None:
             self._type_of_target_ = type_of_target(y)
 
+        # if self.sample_dataset_per_tree and not self.permute_per_tree:
+        #     raise ValueError("sample_dataset_per_tree is only valid when permute_per_tree=True")
+
```
PSSF23: These changes are regressions and should be removed

sktree/stats/forestht.py
```diff
@@ -426,8 +432,7 @@ def test(
         y : ArrayLike of shape (n_samples, n_outputs)
             The target matrix.
         covariate_index : ArrayLike, optional of shape (n_covariates,)
-            The index array of covariates to shuffle, will shuffle all columns by
-            default (corresponding to None).
+            The index array of covariates to shuffle, by default None.
```
PSSF23: These changes are regressions and should be removed

sktree/stats/forestht.py
YuxinB: Okay, adding a check type now.

PSSF23 (commented):
I have corrected the change regressions & formatted the file.

PSSF23 (commented):
@adam2392 @YuxinB The failed checks are about `train_test_samples_`, which takes the argument of stratifier y and not recognized as an attribute in the test. The `property` label was removed. How do you think this should be resolved?

adam2392:
> @adam2392 @YuxinB The failed checks are about `train_test_samples_`, which takes the argument of stratifier y and not recognized as an attribute in the test. The `property` label was removed. How do you think this should be resolved?

I think the issue is that `train_test_samples_` is meant to be a "fitted" property that can be recomputed after-the-fact. However, a reliance on `y` essentially means it is not replicable without access to `y`.

There are a few ideas that come to mind:

- Add `_y` as a cached array: this increases RAM usage and disc usage when picking, but removes any reliance on keeping track of `y`
- Only pass `y` into a private property
- Rename `train_test_samples_` and add a function outside the class `_compute_train_test_samples`, so we don't expose what are the training/testing indices

I am in favor of the first option which is only activated when we need `y` to be stratified (i.e. classification)

adam2392:
I've implemented said changes. @YuxinB you must add yourself to the `contributors.rst` file. You also need to test an additional two if/else statement blocks.

Assuming CIs work, I'll let @sampan501 and @PSSF23 review and merge if they are happy.

examples/hypothesis_testing/plot_MI_gigantic_hypothesis_testing_forest.py
adam2392, line 1: Made example simpler

sktree/stats/forestht.py
```diff
@@ -160,8 +162,31 @@ def reset(self):
...
+                    f"If running on a new dataset, call the 'reset' method."
+                )
+
+            # Type of target should be one that fits a classifier as this is
+            # the only instance where stratification is needed.
+            if (
+                self._type_of_target_ is not None
+                and type_of_target(stratifier) != self._type_of_target_
+            ):
+                raise RuntimeError(
```
adam2392: Please add a test for these lines.

sktree/stats/forestht.py
```diff
@@ -160,8 +162,31 @@ def reset(self):
...
         self._is_fitted = False
         self._seeds = None
+        self._y = None
+
+    def _get_estimators_indices(self, stratifier=None, sample_separate=False):
+        # Check stratifier
+        # if stratifier is None, stratifier is regressor
+        if stratifier is not None:
+            if self._n_samples_ is not None and stratifier.shape[0] != self._n_samples_:
+                raise RuntimeError(
```
adam2392: Please add a test for these lines.

sampan501 (dismissed):
I think the other thing that is missing is documentation for the `stratify` parameter

PSSF23 (commented):
Previously the tests for stratification were not working (they will always pass). I think I corrected them & added new ones @adam2392 requested.

PSSF23 (commented):
EDIT: I redid the tests, as the previous test format can never test stratification.

PSSF23 (commented):
I removed duplicate checks as those would be covered by `check_input` anyway. Explicitly setting the parameter to `False` just to have new checks that check the same thing doesn't make sense to me.

PSSF23 (approved):
@adam2392 I believe the tests should be excluded from codecov? And do you think we need more tests?

adam2392:
Can you add docstring for stratify in the two classes?

PSSF23 (commented):
I set the base forest default value to `False` and removed it from the regressor class init

sampan501:
What set the default of `stratify` to false? Isn't it better to stratify in general according to the sims? Or is it because of the variance in the performance?

adam2392:
Yeah just to give options.

PSSF23:
> What set the default of `stratify` to false? Isn't it better to stratify in general according to the sims? Or is it because of the variance in the performance?

I only set the base forest value to false for regressor, classifier is still default true

PSSF23:
@adam2392 Do you know how to exclude `experimental` and all tests from codecov?

adam2392:
> @adam2392 Do you know how to exclude `experimental` and all tests from codecov?

Experimental we can include, but the tests should be not part of it. I think this might be a bug on codecov cuz it wasn't showing up before.

adam2392:
The build-docs are failing because @YuxinB needs to add herself to the contributors doc: 

```
/home/circleci/project/doc/whats_new/v0.3.rst:18: ERROR: Unknown target name: "yuxin bai".
/home/circleci/project/doc/whats_new/v0.3.rst:28: ERROR: Unknown target name: "yuxin bai".
```

The codecov/project doesn't need to pass

adam2392 (approved):
LGTM once CI green

adam2392:
You'll need to rename the references to the example file, since you changed the name of the example.

```
/home/circleci/project/doc/auto_examples/hypothesis_testing/plot_MI_imbalanced_hyppo_testing.rst:39: WARNING: undefined label: 'sphx_glr_auto_examples_hypothesis_testing_plot_mi_gigantic_hypothesis_testing_forest.py'
```

PSSF23:
> You'll need to rename the references to the example file, since you changed the name of the example.
> 
> ```
> /home/circleci/project/doc/auto_examples/hypothesis_testing/plot_MI_imbalanced_hyppo_testing.rst:39: WARNING: undefined label: 'sphx_glr_auto_examples_hypothesis_testing_plot_mi_gigantic_hypothesis_testing_forest.py'
> ```

Done

adam2392 (approved):

adam2392:
Thanks @YuxinB @PSSF23 and @sampan501

PR #145, opened 2023-10-16
[ENH] Add option to permute per forest fraction

adam2392:
Discussed in Forest meeting today:

- adds a `permute_per_forest_fraction` parameter that permutes the covariate_index a controlled number of times over the entire forest (rather than per tree)
- Helps resolve the humongous usage of RAM when using large `X`, or large forests, or many repeated jobs of `FeatureImportance*Classifier`.

## Before submitting

## After submitting


adam2392:
I'll let #143 get merged first, so I can test out changes with this new functionality

PSSF23 (commented):
The tests didn't pass because some indices were not stratified during testing. I also feel the `samples` variable is used too many times, which could cause confusion. In classifier `_statistic`, `samples` could mean test indices or non-nan indices.

PSSF23:
Should we close this?

adam2392:
It's a simple addition that's backwards compatible so I think I can just finish adding it.

adam2392:
Is this mergable? @PSSF23 to preserve the old behavior, just set `permute_forest_fraction = 1.0 / n_estimators`.

PSSF23:
@adam2392 We should keep the same behavior for the `permute_per_tree` parameter, so Sam would not need to change his code. The permute fraction should default to each tree with `None` value, similar to how sklearn handles the relationship between `max_samples` and `bootstrap`. Like raising an error if fraction is not `None` when `permute_per_tree` is `False`.

adam2392:
permute_per_tree default for Sam is `False` corresponding to no permutation per tree.

permute_per_forest_fraction = None would have the same default. Is this what you mean?

adam2392:
> @adam2392 We should keep the same behavior for the `permute_per_tree` parameter, so Sam would not need to change his code. The permute fraction should default to each tree with `None` value, similar to how sklearn handles the relationship between `max_samples` and `bootstrap`. Like raising an error if fraction is not `None` when `permute_per_tree` is `False`.

Currently, permute_per_tree default for Sam is `False` corresponding to no permutation per tree.

With this PR, permute_per_forest_fraction = None would have the same default. Is this what you mean?

PSSF23 (approved):
@adam2392 Actually it might not be relevant to Sam at the current stage of implementation as he's only running statistics. 

My original thought is to preserve any current use of `permute_per_tree` and treat `permute_per_forest_fraction` as add-on to it. But as the new parameter fully covers the original use, I think it's fine to upgrade. Can you fix the error in the regressor test?

PR #152, opened 2023-10-25
[ENH] Allow sampling feature sets separately in MultiViewDTC

### LINKED ISSUE #151: Add parameter to multi-view splitter to add `non_uniform_sampling`
Currently, max_features works by computing the number of m-try over `n_features` and then randomly samples from each feature set without replacement one feature-set at a time until max_feature is reached.

This may not be desirable because it doesn't sample each feature-set with the same ratio. E.g.  if there are max_features=20 for feature-set A with (`n_features_set_A = 10`) and feature-set B with (`n_features_set_B = 1000`), then feature-set A is always sampled fully, while feature-set B is only sampled at a rate of 10/1000.

Instead, we can apply `non_uniform_sampling` and `sqrt`, we can have max_features be an array that samples sqrt(10) features randomly in feature-set A and sqrt(1000) in feature-set B. This way, the same scaling on the feature-sampling is applied.

adam2392:
- Allows feature-sets to be sampled with the same scaling factor when `max_features` is a float, 'sqrt', or 'log'
- Adds an experiment demonstrating multi-view superiority on a set of simulations when the feature-size of the second view in a 2-view dataset increases. All other simulation parameters are held constant (`examples/hypothesis_testing/plot_co_MIGHT_alternative.py` and `examples/hypothesis_testing/plot_co_MIGHT_null.py`)

The main files to take a look at are:

- sktree/stats/forestht.py
- sktree/tree/_multiview.py
- sktree/tree/_oblique_splitter.pxd
- sktree/tree/_oblique_splitter.pyx

The rest are cosmetic fixes, or small documentation issues I noticed.

## Before submitting

## After submitting


sktree/tests/test_honest_forest.py
```diff
@@ -252,3 +255,36 @@ def test_importances(dtype, criterion):
...
+
+    # XXX: surprisingly, when we use the default which uses the fork DecisionTree,
+    # we get different results
+    # clf = HonestForestClassifier(n_estimators=10, random_state=0)
+    # honest_scores = cross_val_score(clf, X, y, cv=5)
+    # print(honest_scores)
+
+    print(honestsk_scores, honest_scores)
+    print(np.mean(honestsk_scores), np.mean(honest_scores))
+    assert_allclose(np.mean(honestsk_scores), np.mean(honest_scores))
```
adam2392: @sampan501 it seems there is some difference in the randomness when we pass in an explicit class vs letting the default handle itself in `HonestForestClassifier`. When you ran quadratic exps, which one did you do?

sampan501: I explicitly passed in `DecisionTreeClassifier()` like you have written above

adam2392: cc: @PSSF23 

In this branch, I basically test:

- sklearn tree vs sklearn-fork tree
- honest tree wrapping sklearn tree vs sklearn-fork tree
- honest forest wrapping sklearn tree vs sklearn-fork tree

In this specific file is the honest forest. We basically get exactly the same performance up to machine precision in trees and almost identical performance in honest forest, so it's confusing me. I think we need a small-scale version of Sambit's power simulation to unit-test and reproduce the performance regression because I'm not seeing any.

adam2392: @sampan501 how many repetitions were you using to compare power curves of the quadratic w/ sklearn-tree vs sklearn-fork tree?

sampan501: I was using about 100 for both. I'm getting consistently higher power using the old method compared to using the `scikit-tree` methods

adam2392 (commented):
The main files to take a look at are:

- sktree/stats/forestht.py
- sktree/tree/_multiview.py
- sktree/tree/_oblique_splitter.pxd
- sktree/tree/_oblique_splitter.pyx

The rest are cosmetic fixes, or small documentation issues I noticed.

sktree/stats/forestht.py
```diff
@@ -575,14 +579,14 @@ def test(
...
 
         # metric^\pi_j - metric_j, which is centered at 0
         null_dist = metric_star_pi - metric_star
 
         # compute pvalue
         if metric in POSITIVE_METRICS:
-            pvalue = (1 + (null_dist <= observe_stat).sum()) / (1 + n_repeats)
+            pvalue = (1 + (null_dist <= observe_test_stat).sum()) / (1 + n_repeats)
         else:
             pvalue = (1 + (null_dist >= observe_stat).sum()) / (1 + n_repeats)
```
PSSF23: The variable name is not fixed here.

PSSF23: Also the return value at line 600.

adam2392: Oh good catch. Addressed.

examples/hypothesis_testing/plot_MI_imbalanced_hyppo_testing.py
```diff
@@ -91,18 +91,19 @@ def make_multiview_classification(
 
 
 n_samples = 100
-n_features = 10000
 n_features_1 = 10
-n_features_2 = n_features
+n_features_2 = 10_000
```
sampan501, line 95: Why does this have a underscore?

sktree/tests/test_honest_forest.py
adam2392: This is now re-filed as an issue in #171 and some additional tests were added in #164

sktree/tree/_multiview.py
```diff
@@ -360,6 +378,51 @@ def _build_tree(
...
+                        f"the number of features in feature set {idx}: {n_features_in_}, but "
+                        f"max_features = {max_features} when applying sampling"
+                        f"per feature set."
+                    )
+
+                max_features_per_set.append(max_features)
+            self.max_features_ = np.sum(max_features_per_set)
+            if self.max_features_ > n_features:
+                raise ValueError(
+                    "max_features is greater than the number of features. "
```
sampan501: I think the error trace should write out the actual values of `max_features` and `n_features` to be more useful

sampan501 (approved):
Few minor things, but looks good to me

examples/hypothesis_testing/plot_MI_imbalanced_hyppo_testing.py
adam2392: To make it easier to read that it is `10_000` and remove the unnecessary variable `n_features`

sktree/tree/_multiview.py
adam2392: Added.

PSSF23 (approved):

PR #161, opened 2023-10-31
FIX improve performance by updating max_features

PSSF23:


PSSF23 (commented):
@adam2392 this change should improve the mv performance, which is too low in the current example.

adam2392:
Can you paste the rendered artifact link here? I can take a look on my phone.

PSSF23:
> Can you paste the rendered artifact link here? I can take a look on my phone.

I don't know how to access it, but the local performance is 0.64/0.57, definitely better than what we have now.

adam2392:
For future ref: under the GH-actions: `build_docs` links to the circleCI page and `Artifacts` uploaded has the html pages

adam2392 (approved):

PR #175, opened 2023-11-17
FIX correct posterior predictions for PFC

PSSF23:
- Add posterior predictions for appropriate metrics (MI, AUC...)
- Correct permutation, remove permutation on test set


adam2392 (commented):
Can you add a short test, or modify an existing one to add the new metrics that we want to test? 

You'll also need to run the style formatters

adam2392:
Apologies! I messed up the main branch, and as a result this PR auto-closed. Do you mind resubmitting it if it's of interest?

PR #192, opened 2023-12-20
[Paper] Add function to calculate metrics from posteriors

SUKI-O:
Add function to calculate metrics from posteriors from saved runs in order to avoid re-running the entire code for different metrics. 

-

## Before submitting

## After submitting

### DIFF
--- a/sktree/tests/test_utils.py
+++ b/sktree/tests/test_utils.py
@@ -1,0 +1,71 @@
+import numpy as np
+import pytest
+from sklearn import datasets
+
+from sklearn.ensemble import RandomForestClassifier
+
+from ..stats.utils import _SAS98, compute_posterior_metrics
+
+# set the constant seed for reproducibility
+SEED = 123
+np.random.seed(SEED)
+# generate toy samples of posteriors, targets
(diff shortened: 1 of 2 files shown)


sktree/tests/test_utils.py
adam2392, line 1: Here I would test with 

- `FeatureImportanceClassifier` with the HonestForest setup we have (i.e. MIGHT, or CoMIGHT)
- Something else like SVM from sklearn

Note that posteriors returned from something like sklearn has a diff shape.

adam2392: You can test against metric functions that take in "y_score":

- http://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html
- https://scikit-learn.org/stable/modules/generated/sklearn.metrics.top_k_accuracy_score.html#sklearn.metrics.top_k_accuracy_score
- our internal implementation of mutual information

A positive test is just a "smoke test" to make sure it runs and the value is "expected". Then a negative test could be where you change something that you can foresee being a common error-surface in our workflow and make sure we error-check properly.

PR #199, opened 2024-01-16
ENH add MI simulations

PSSF23:
Add simulations that use conditional entropy to calculate MI


sktree/datasets/hyppo.py
```diff
@@ -43,3 +45,279 @@ def make_quadratic_classification(n_samples: int, n_features: int, noise=False,
...
+
+
+def make_trunk_classification(
+    n_samples,
+    n_dim=10,
+    m_factor: int = -1,
+    rho: int = 0,
+    band_type: str = "ma",
+    return_params: bool = False,
+    mix: int = 0,
```
adam2392, line 57:
```suggestion
    mix: bool = False,
```

sktree/datasets/hyppo.py
```diff
@@ -43,3 +45,279 @@ def make_quadratic_classification(n_samples: int, n_features: int, noise=False,
...
+        distribution to obtain the mean-vector of the second distribution.
+        By default -1.
+    rho : float, optional
+        The covariance value of the bands. By default 0 indicating, an identity matrix is used.
+    band_type : str
+        The band type to use. For details, see Example 1 and 2 in :footcite:`Bickel_2008`.
+        Either 'ma', or 'ar'.
+    return_params : bool, optional
+        Whether or not to return the distribution parameters of the classes normal distributions.
+    mix : int, optional
```
adam2392:
```suggestion
    mix : bool, optional
```

adam2392 (approved):
Otw LGTM. I have no idea why the changelog check is failing, but we can bypass it. Same w/ style

sampan501:
@adam2392 `mix` is currently an int, and realizing now should be a float, that takes values between 0 and 1. I don't think it should be a bool.

adam2392:
Good point. Wanna add that and an error check on the value of mix being in (0,1) there?

sktree/datasets/hyppo.py
```diff
@@ -43,3 +45,297 @@ def make_quadratic_classification(n_samples: int, n_features: int, noise=False,
...
+    covs : list of ArrayLike of shape (n_dim, n_dim), dtype=np.float64
+        The covariance for each class. Returned if ``return_params`` is True.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+    rng = np.random.default_rng(seed=seed)
+
+    mu_1 = np.array([1 / np.sqrt(i) for i in range(1, n_dim + 1)])
```
PSSF23, line 113: Aren't these `mu`'s reversed from what we defined on board? F should be `mu` and G should be `-mu`. @sampan501 @adam2392

adam2392: I think it should be fine since it is symmetrical.

PSSF23: I don't think we should add extra confusion in it.

sampan501: I personally don't feel strongly either way

PSSF23: ok

adam2392:
Thanks @PSSF23 and @sampan501 ! Putting this in, so we can move towards some of the other tasks

If the mu0 vs mu1 ends up being confusing, feel free to propose the change in another PR.

PR #211, opened 2024-01-31
[ENH] A modular approach to constructing forests that have permuted covariates

adam2392:
This may eventually replace `FeatureImportance*` if we need it to.

- We separate the training of a permuted forest and the final computation of a pvalue
- I will also enable the permutation of covariates within Cython if the power computation experiment determines bootstrapping is better.
- Adds `oob_samples_` to any forest that is fit with `bootstrap` on

The forests should be classes without too much logic addition and the pvalue computation logic can be wrapped into a function.

## Before submitting

## After submitting


PSSF23 (approved):
LGTM if checks all pass.

PR #218, opened 2024-02-13
ENH change the default parameters and statistics of honest forests

PSSF23:
- Update the default values for `max_features`, `max_samples`, `bootstrap`, `stratify`, and statistic metric.


adam2392:
I think this breaks some of the unit-tests, but idr if this means the tests should be refactored, or if this introduces certain edge case bugs.

PR #227, opened 2024-02-21
MAINT, API Change trunk-classification into three separate functions for generating trunk, trunk-mix, trunk-overlap and marron-wand

adam2392:
- separates `make_trunk_classification` into three functions: `make_trunk_class*`, `make_trunk_mixture_class*`, and `make_marron_wand_class*` so that we can separate logic for simulating trunk, trunk-overlap, trunk-mix and marron-wand simulations

## Before submitting

## After submitting

### DIFF
--- a/sktree/datasets/hyppo.py
+++ b/sktree/datasets/hyppo.py
@@ -76,7 +76,8 @@ def make_trunk_classification(
     n_dim=4096,
     n_informative=256,
     simulation: str = "trunk",
-    m_factor: int = -1,
+    mu_0: int = 0,
+    mu_1: int = 1,
     rho: int = 0,
     band_type: str = "ma",
     return_params: bool = False,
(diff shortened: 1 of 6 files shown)


sktree/datasets/hyppo.py
```diff
@@ -205,7 +213,7 @@ def make_trunk_classification(
             )
         )
     elif simulation == "trunk_mix":
-        mixture_idx = rng.choice(2, n_samples // 2, replace=True, shuffle=True, p=[mix, 1 - mix])
+        mixture_idx = rng.choice(2, n_samples // 2, replace=True, shuffle=True, p=[mix, 1 - mix])  # type: ignore
         norm_params = [[mu_0, cov * (2 / 3) ** 2], [mu_1, cov * (2 / 3) ** 2]]
```
adam2392: @sampan501 can you explain why we have a (2/3)**2 here and only in trunk-mix?

sampan501: When variance is 1, trunk-mix does not look bimodal at low dimensions. I set it to (2/3)**2 since that is consistent with Marron and Wand bimodal

sampan501 (approved):

YuxinB (approved):

sampan501:
@adam2392 Good to merge after changelog update

sktree/datasets/hyppo.py
```diff
@@ -106,19 +99,179 @@ def make_trunk_classification(
...
+            rng.multivariate_normal(
+                np.zeros(n_informative), cov, n_samples // 2, method=mvg_sampling_method
+            ),
+            (1 - w)
+            * rng.multivariate_normal(
+                np.zeros(n_informative), cov, n_samples // 2, method=mvg_sampling_method
+            )
+            + w * G.reshape(n_samples // 2, n_informative),
+        )
+    )
```
adam2392: @sampan501 can you explain what the role of `w` is? I'm not sure why it's passed via `mu_1`, or if it should be?

YuxinB: From this code, if I understand right, w should be the probability that we sample from G. Like (1-w)*N(0,1)+w*G, G is any Marron-Wand simulation.

sampan501 (changes requested):
Looks good! For simplicity, there is a lot of error checks that are common for all 3 functions. Maybe we can wrap those up into a private method?

sampan501:
Other than that, pending CI, lgtm

adam2392:
> Looks good! For simplicity, there is a lot of error checks that are common for all 3 functions. Maybe we can wrap those up into a private method?

Good idea. Rn tho there's only a few.  We could migrate the covariance matrix stuff to a `_default_covariance(...)` which handles the check and setting of covariance possibly. But maybe we can leave for later depending on any future added complexity. Otw over-engineering maybe :p

sampan501:
Fair, I'll make an issue for it

sampan501 (approved):

SUKI-O (approved):

PR #234, opened 2024-02-27
[ENH, FIX] Allow scaling_factor in trunk_mixture_classification to be specified

adam2392:
- The covariance should be able to be specified, such that it is the identity matrix in trunk mixture.

## Before submitting

## After submitting

### DIFF
--- a/sktree/datasets/hyppo.py
+++ b/sktree/datasets/hyppo.py
@@ -1,5 +1,3 @@
-from typing import Optional
-
 import numpy as np
 from scipy.integrate import nquad
 from scipy.stats import entropy, multivariate_normal
@@ -79,6 +77,7 @@ def make_marron_wand_classification(
     rho: int = 0,
     band_type: str = "ma",
     return_params: bool = False,
+    scaling_factor: float = 1.0,
     seed=None,
 ):
     """Generate Marron-Wand binary classification dataset.
@@ -120,6 +119,8 @@ def make_marron_wand_classification(
         Either 'ma', or 'ar'.
     return_params : bool, optional
         Whether or not to return the distribution parameters of the classes normal distributions.
+    scaling_factor : float, optional
+        The scaling factor for the covariance matrix. By default 1.
     seed : int, optional
         Random seed, by default None.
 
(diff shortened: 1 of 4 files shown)


PSSF23 (commented):
Why is 2/3 default? We don't have any simulations that use such scaling right now. Maybe in the future, but I would set the default to be identity.

sktree/datasets/hyppo.py
```diff
@@ -281,6 +284,9 @@ def make_trunk_mixture_classification(
     mix : int, optional
         The probabilities associated with the mixture of Gaussians in the ``trunk-mix`` simulation.
         By default 0.5.
+    scaling_factor : float, optional
+        The scaling factor for the covariance matrix. By default None, which
+        defaults to (2/3)**2.
```
sampan501: I think should be 1 by default

adam2392:
> Why is 2/3 default? We don't have any simulations that use such scaling right now. Maybe in the future, but I would set the default to be identity.

cc: @sampan501 what was the reason again? Iforgot.

sktree/datasets/hyppo.py
```diff
@@ -236,6 +238,7 @@ def make_trunk_mixture_classification(
     band_type: str = "ma",
     return_params: bool = False,
     mix: float = 0.5,
+    scaling_factor: Optional[float] = None,
```
sampan501: Thoughts on putting this for all 3 simulation functions? The default scaling factor for the marron and wand can be whatever that simulations' factor currently is

sampan501 (changes requested):

sampan501:
> > Why is 2/3 default? We don't have any simulations that use such scaling right now. Maybe in the future, but I would set the default to be identity.
> 
> cc: @sampan501 what was the reason again? Iforgot.

I had done that before because when making the visualization, using 1 doesn't look bimodal. So, I stuck with the Marron and Wand default. For the sake of this though, I think we should just use 1

sktree/datasets/hyppo.py
adam2392: Done.

sktree/datasets/hyppo.py
adam2392: Okay sure changed.

sampan501 (approved):
Pending green CIs, lgtm

PR #236, opened 2024-02-28
bug fixes for trunk sims

sampan501:
Fixes #

- Bug fix: When increasing the dimensions, we don’t get the same values given the same seed.
- Doc fix: Odd values for n_samples give 1 values less than they should
- Bug fix: Values for trunk are the same for the first group but different for the second group

## Before submitting

## After submitting

### DIFF
--- a/sktree/datasets/hyppo.py
+++ b/sktree/datasets/hyppo.py
@@ -197,7 +197,10 @@ def make_marron_wand_classification(
     )
     # the parameters used for each Gaussian in the mixture for each Marron Wand simulation
     norm_params = MarronWandSims(n_dim=n_informative, cov=cov)(simulation)
-    rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    if seed:
+        rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    else:
+        rngs = [rng for _ in range(n_samples // 2)]
     G = np.fromiter(
         (
             rngs[i].multivariate_normal(
@@ -211,7 +214,10 @@ def make_marron_wand_classification(
     # as the dimensionality of the simulations increasing, we are adding more and
     # more noise to the data using the w parameter
     w_vec = np.array([1.0 / np.sqrt(i) for i in range(1, n_informative + 1)])
-    rngs_F = [np.random.default_rng(seed=seed + i) for i in range(2)]
+    if seed:
+        rngs_F = [np.random.default_rng(seed=seed + i) for i in range(2)]
+    else:
+        rngs_F = [rng, rng]
     X = np.vstack(
         (
             rngs_F[0].multivariate_normal(
@@ -369,7 +375,10 @@ def make_trunk_mixture_classification(
         method = "svd"
 
     mixture_idx = rng.choice(2, n_samples // 2, replace=True, shuffle=True, p=[mix, 1 - mix])  # type: ignore
-    rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    if seed:
+        rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    else:
+        rngs = [rng for _ in range(n_samples // 2)]
 
     norm_params = [[mu_0_vec, cov], [mu_1_vec, cov]]
     X_mixture = np.fromiter(
--- a/sktree/datasets/hyppo.py
+++ b/sktree/datasets/hyppo.py
@@ -197,16 +197,10 @@ def make_marron_wand_classification(
     )
     # the parameters used for each Gaussian in the mixture for each Marron Wand simulation
     norm_params = MarronWandSims(n_dim=n_informative, cov=cov)(simulation)
-    if seed:
-        rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
-    else:
-        rngs = [rng for _ in range(n_samples // 2)]
     G = np.fromiter(
         (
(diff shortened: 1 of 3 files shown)


sktree/datasets/hyppo.py
```diff
@@ -96,7 +96,7 @@ class distribution approaches the first class distribution by a factor of :math:
     Parameters
     ----------
     n_samples : int
-        Number of sample to generate.
+        Number of sample to generate. Must be an even number.
```
adam2392:
```suggestion
        Number of sample to generate. Must be an even number, else the total number of samples generated will be ``n_samples - 1``
```

Right?

sktree/datasets/hyppo.py
```diff
@@ -197,31 +197,41 @@ class distribution approaches the first class distribution by a factor of :math:
     )
     # the parameters used for each Gaussian in the mixture for each Marron Wand simulation
     norm_params = MarronWandSims(n_dim=n_informative, cov=cov)(simulation)
+    if seed:
+        rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    else:
+        rngs = [rng for _ in range(n_samples // 2)]
```
adam2392: Do we not need to also re-seed here? or is `None` handle that?

sktree/datasets/hyppo.py
```diff
@@ -265,7 +275,7 @@ def make_trunk_mixture_classification(
     Parameters
     ----------
     n_samples : int
-        Number of sample to generate.
+        Number of sample to generate. Must be an even number.
```
adam2392: Same comment as above

sktree/datasets/hyppo.py
```diff
@@ -365,22 +375,31 @@ def make_trunk_mixture_classification(
         method = "svd"
 
     mixture_idx = rng.choice(2, n_samples // 2, replace=True, shuffle=True, p=[mix, 1 - mix])  # type: ignore
+    if seed:
+        rngs = [np.random.default_rng(seed=seed + i) for i in range(n_samples // 2)]
+    else:
+        rngs = [rng for _ in range(n_samples // 2)]
```
adam2392: Same question as above

adam2392 (commented):
Nice! Few minor changes and one question and then LGTM

sktree/datasets/hyppo.py
sampan501: yes

sktree/datasets/hyppo.py
sampan501: I think we only care about this when we fix a seed right?

sktree/datasets/hyppo.py
sampan501: https://github.com/neurodata/scikit-tree/pull/236/commits/02af22dcdde829ee52812b9f1f5f6462201281a8

sktree/datasets/hyppo.py
adam2392: Oh I see. Because if we don't fix the seed then each run will be different anyways?

adam2392 (approved):

sktree/datasets/hyppo.py
sampan501: yep

sampan501:
Found some bugs, I'll squash before I ask for another review

sampan501:
Ok, I believe everything should be correct now. tldr of the changes:

1. Unit tests were testing if the same array is equal since the second array was just `n_samples + 1`. Now, it tests if group 1 elements are the same, group 2 elements are the same, and the noise is the same
2. Was defining a new seed and iterating by an `int`. Switched to the `rng.spawn` method instead
3. Ocassionally, I had to reinitalize a new variable with the same generator `int`. This I learned from validating unit tests.

PSSF23 (approved):

sktree/datasets/hyppo.py
```diff
@@ -199,36 +199,47 @@ class distribution approaches the first class distribution by a factor of :math:
...
+            rng_children.multivariate_normal(*(norm_params[i]), size=1, method=mvg_sampling_method)
+            for i, rng_children in zip(mixture_idx, rng.spawn(n_samples // 2))
         ),
         dtype=np.dtype((float, n_informative)),
     )
 
     # as the dimensionality of the simulations increasing, we are adding more and
     # more noise to the data using the w parameter
     w_vec = np.array([1.0 / np.sqrt(i) for i in range(1, n_informative + 1)])
+    rng_F = np.random.default_rng(seed=seed).spawn(2)
```
adam2392, line 213: Why not use the existing rng to spawn 2 children here vs re defining a rng for F?

Could be useful to inline comment

adam2392 (approved):
To confirm the separate rng instances are required to keep noise variables the same even if n informative changes? 

Would be nice to document that just as in line comments for anyone who reads it in the future. 

Otw LGTM!

sktree/datasets/hyppo.py
sampan501: I'll comment. It was from when I was evaluating the unit tests; had the same issue of continue to sample from `rng` leading to failed tests

sampan501:
I'll quickly verify that simulations look correct in scatter plots, add the comment, and then merge

sktree/datasets/hyppo.py
sampan501: https://github.com/neurodata/scikit-tree/pull/236/commits/d522ea9e2e7989763544e012ed6dbbd70ede4492

sampan501:
Ok verified! I'll merge when CIs are green

PR #254, opened 2024-04-15
FIX correct simulation class label for linear

PSSF23:
- The class labels are reversed for the linear simulations


PSSF23 (commented):
Reminder to self: change tutorial after correction is in `pip`

adam2392 (approved):
Okay, I'll let you merge. Wanna also change the downstream PR to account for this?

PSSF23 (commented):
Yes, we need to modify the tutorials and other possible problems. But only after it's in `pip`

PR #256, opened 2024-04-16
ENH add treeple tutorials to website

PSSF23:
Add treeple tutorials for:

- single-view MI, pAUC, S@98, p-value
- multi-view CMI, S@98, p-value
- gaussian mixture model truth

Open to suggestions and feedback. I realized that some content overlaps with existing tutorials, which could be outdated and should be integrated/removed.


PSSF23 (commented):
The tutorial simulation needs to be modified after #254 is merged and the change is pushed to `pip`.

examples/treeple/treeple_tutorial_pvalue.ipynb
YuxinB, line 1: for the p-val with Coleman's method, do we need to add something that we are permitting trees from two forests at "Permute Tree Posterior"?

PSSF23: Permuting tree posteriors and trees are equivalent in our case. Which part of the text you think could be optimized?

YuxinB (approved):

examples/treeple/treeple_tutorial_pvalue.ipynb
YuxinB: Yeah, my concern is just the "Permute Tree Posterior" might be a little confusing for readers? Can we add something like your statement that "Permuting tree posteriors and trees are equivalent in our case"?

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
...
+#
+# Given samples from ``X`` and ``Y``, the independent hypothesis and its
+# alternative are stated as:
+#
+# .. math:: H_0 : F_{XY} = F_X F_Y
+#
+# .. math:: H_A : F_{XY} \neq F_X F_Y
+#
+# By computing the p-value using ``treeple``, we can test if :math:`H_0`
+# would be rejected, which confirms the joint distribution. The p-vlaue is
```
sampan501: which would confirm that X and Y are not independent. Also, p-value is spelled wrong

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
...
+# generated by comparing the observed statistic difference with permuted
+# differences, using mutual information as an example.
```
sampan501: Would write something like the p-value is the probability of observing a statistic more extreme than the null. One example of a statistic we can compute is mutual information (MI)

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
...
+# --
+#
+# Mutual Information (*MI*) measures the mutual dependence between *X* and
+# *Y*. It can be calculated by the difference between the class entropy
+# (``H(Y)``) and the conditional entropy (``H(Y | X)``):
+#
+# .. math:: I(X; Y) = H(Y) - H(Y\mid X)
+#
+# With a binary class simulation as an example, this tutorial will show
+# how to use ``treeple`` to use the statistic and test the hypothesis.
```
sampan501: to compute the statistic and p-value

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
...
+)
+
+# fit the model and obtain the tree posteriors
+_, observe_proba = build_hyppo_oob_forest(est, X, y)
+
+# generate forest posteriors for the two classes
+observe_proba = np.nanmean(observe_proba, axis=0)
+
+
+# scatter plot the posterior probabilities for class one
```
sampan501: write something about observed

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
...
+# ---------------------
+
+
+pvalue = (1 + (mix_diff >= observed_diff).sum()) / (1 + PERMUTE)
+
+print("p-value is:", round(pvalue, 2))
+if pvalue < 0.05:
+    print("The null hypothesis is rejected.")
+else:
+    print("The null hypothesis is not rejected.")
```
sampan501: This is very informative about how to compute a p-value, but not sure if we need all of this. I think that point of the tutorial is to show the user how to compute a p-value using the functions already in `treeple`

sampan501 (changes requested):

PSSF23 (commented):
TODO: update p-value calculation with native function.

examples/a_treeple/README.txt
```diff
@@ -0,0 +1,6 @@
+.. _treeple:
+
+Treeple
```
adam2392: Perhaps this should be labeled how to use treeple for hypothesis testing.

examples/a_treeple/treeple_tutorial_CMI.py
```diff
@@ -0,0 +1,149 @@
+"""
+====================================
+Treeple tutorial for calculating CMI
+====================================
+"""
```
adam2392: We should add a short paragraph describing how we actually do the estimate. And maybe what we will show. It will help users acquaint with the example. 

TODO: also link reference to paper when we make preprint online.

examples/a_treeple/treeple_tutorial_CMI.py
```diff
@@ -0,0 +1,149 @@
...
+import numpy as np
+from scipy.stats import entropy
+
+from sktree.datasets import make_trunk_classification
+from sktree.ensemble import HonestForestClassifier
+from sktree.stats import build_hyppo_oob_forest
+from sktree.tree import MultiViewDecisionTreeClassifier
+
+# %%
+# CMI
```
adam2392: I would spell out CMI first. Not everyone knows what CMI stands for. In addition, perhaps it is useful to specify in words what CMI will tell us.

examples/a_treeple/treeple_tutorial_CMI.py
```diff
@@ -0,0 +1,149 @@
...
+plt.legend()
+plt.show()
+
+
+# %%
+# Calculate the statistic
+# -----------------------
+
+
+def Calculate_MI(y_true, y_pred_proba):
```
adam2392: Is it worth making the existing function we have public so you don't need to redefine one here?

examples/a_treeple/README.txt
```diff
@@ -0,0 +1,6 @@
...
+-------
+
+Examples concerning how to use treeple as hypothesis test tools.
```
adam2392: Perhaps we can also specify an order for the tutorials. Namely posterior estimates seems like a logical first choice? Wdyt?

examples/a_treeple/treeple_tutorial_GMM.py
```diff
@@ -0,0 +1,177 @@
+"""
+============================================================
+Treeple tutorial for estimating true posteriors & statistics
```
adam2392: Same comment here regarding adding a paragraph describing what will be shown.

examples/a_treeple/treeple_tutorial_GMM.py
```diff
@@ -0,0 +1,177 @@
...
+        color="r",
+        alpha=0.6,
+        label="pAUC@90 = " + str(round(pAUC, 2)),
+        linestyle="--",
+    )
+    plt.legend()
+    return pAUC
+
+
+pAUC = Calculate_pAUC(y, pos, max_fpr=0.1)
```
adam2392: Do we want to print it out?

examples/a_treeple/treeple_tutorial_GMM.py
```diff
@@ -0,0 +1,177 @@
...
+    else:
+        fpr, tpr, thresholds = roc_curve(
+            y_true, y_pred_proba[:, 1], pos_label=2, drop_intermediate=False
+        )
+    RocCurveDisplay(fpr=fpr, tpr=tpr).plot(label="ROC Curve")
+    # Calculate pAUC at the specific threshold
+    pAUC = roc_auc_score(y_true, y_pred_proba[:, 1], max_fpr=max_fpr)
+
+    pos = np.where(fpr == max_fpr)[0][-1]
+    plt.fill_between(
```
adam2392: Do we want to separate plotting and obtaining the statistic? Makes it a bit cleaner to understand from a user perspective

examples/a_treeple/treeple_tutorial_MI.py
```diff
@@ -0,0 +1,97 @@
+"""
+===================================
+Treeple tutorial for calculating MI
```
adam2392: Same comment here regarding adding a paragraph describing what will be shown. 

Also spell out MI before using abbreviation.

examples/a_treeple/treeple_tutorial_SA98.py
```diff
@@ -0,0 +1,115 @@
+"""
+=====================================
+Treeple tutorial for calculating S@98
```
adam2392: Same comment here regarding adding a paragraph describing what will be shown. 

Also spell out what S@98 is. Or to make this more general, state this is S@k-specificity and we will specifically look at S@98.

examples/a_treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,186 @@
+"""
+========================================
+Treeple tutorial for calculating p-value
```
adam2392: What is the pvalue of? I.e. what is the null and alternative.

examples/a_treeple/treeple_tutorial_pvalue.py
adam2392: Yes I would just include a sentence in the very top of the example stating how to interpret the pvalue and then remove these lines of code

adam2392 (changes requested):
Overall great start!

I think the summary is:

- Let's try to tighten up how each example is presented to the user. Assume the user knows very little. I.e. abbreviations should be defined. Basic concepts should be defined. 
- We also maybe want to consider allowing metrics to be public functions so you can just import them? Assuming all public functions either take in a 2D posterior, this is very similar to scikitlearns metrics API. If we need the public functions to take a 3D posterior over trees that's fine too.

examples/a_treeple/treeple_tutorial_CMI.py
PSSF23: That would be best.

examples/a_treeple/treeple_tutorial_GMM.py
PSSF23: It's in the figure legend, so not necessarily.

examples/a_treeple/treeple_tutorial_GMM.py
PSSF23: It would be best if the local function could be hidden. But sphinx doesn't let me have a separate script file.

examples/treeple/treeple_tutorial_pvalue_multiview.py
```diff
@@ -0,0 +1,238 @@
...
+#
+# .. math:: H_0 : F_{X,Y|Z} = F_{X|Z} F_{Y|Z}
+#
+# .. math:: H_A : F_{X,Y|Z} \neq F_{X|Z} F_{Y|Z}
+#
+# By computing the p-value using ``treeple``, we can test if :math:`H_0`
+# would be rejected, which confirms that ``X|Z`` and ``Y|Z`` are not independent.
+# The p-value is generated by comparing the observed
+# statistic difference with permuted differences, using conditional mutual
+# information as an example.
```
adam2392:
```suggestion
# information as a test statistic in this example.
```

examples/treeple/treeple_tutorial_GMM.py
```diff
@@ -0,0 +1,177 @@
...
+
+warnings.filterwarnings("ignore")
+
+# %%
+# True posterior estimation
+# -------------------------
+#
+# As we know the true priors of each class, we can generate a sufficient
+# amount of samples to estimate the true posteriors and corresponding
+# statistics like *MI*, *pAUC*, and *S@98*.
```
adam2392, line 24:
```suggestion
# In many simulation settings, we are interested in knowing the true statistics. Sometimes these statistics
# are not analytically computable, so we demonstrate how to approximate these numerically. Given enough
# samples, these numerical approximations converge to the true statistics.
# 
# As we know the true priors of each class, we can generate a sufficient
# amount of samples to estimate the true posteriors and corresponding
# statistics like *MI*, *pAUC*, and *S@98*.
```

examples/treeple/treeple_tutorial_pAUC.py
```diff
@@ -0,0 +1,119 @@
...
+# ------
+#
+# Partial area under the ROC curve (*pAUC*) integrates the true positive
+# rates (*TPR*) when the false positive rates (*FPR*) are below a specific
+# percentage threshold. Then the value is normalized by that percentage.
+#
+# .. math:: pAUC@r = \frac{100}{100 - r} \int_{T_r}^\infty \int_{\mathcal{X}} \mathbb{I}\{\eta(X_1) > \eta(X_0) \} dF_1 dF_0
+#
+# With a binary class simulation as an example, this tutorial will show
+# how to use ``treeple`` to calculate the statistic with 90% specificity
```
adam2392, line 26:
```suggestion
# how to use ``treeple`` to calculate the AUC statistic at a specified
# 90% specificity
```

examples/treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,189 @@
...
+#
+# .. math:: H_0 : F_{XY} = F_X F_Y
+#
+# .. math:: H_A : F_{XY} \neq F_X F_Y
+#
+# P-value is the probability of observing a statistic more extreme than the null.
+# By computing the p-value using ``treeple``, we can test if :math:`H_0`
+# would be rejected, which confirms that X and Y are not independent. The p-value is
+# generated by comparing the observed statistic difference with permuted
+# differences, using mutual information as an example.
```
adam2392, line 30:
```suggestion
# differences, using mutual information as the test statistic in this example.
```

examples/treeple/treeple_tutorial_pvalue_multiview.py
```diff
@@ -0,0 +1,238 @@
...
+# ---------------------
+
+
+pvalue = (1 + (mix_diff >= observed_diff).sum()) / (1 + PERMUTE)
+
+print("p-value is:", round(pvalue, 2))
+if pvalue < 0.05:
+    print("The null hypothesis is rejected.")
+else:
+    print("The null hypothesis is not rejected.")
```
adam2392, line 238: Should we remove these LOC? https://github.com/neurodata/scikit-tree/pull/256/files/2c487755d8a0827b7344b2bee6da8eabef1d400e#r1567919332

examples/treeple/treeple_tutorial_pvalue.py
```diff
@@ -0,0 +1,189 @@
...
+# ---------------------
+
+
+pvalue = (1 + (mix_diff >= observed_diff).sum()) / (1 + PERMUTE)
+
+print("p-value is:", round(pvalue, 2))
+if pvalue < 0.05:
+    print("The null hypothesis is rejected.")
+else:
+    print("The null hypothesis is not rejected.")
```
adam2392, line 189: Should we remove these LOC? https://github.com/neurodata/scikit-tree/pull/256/files/2c487755d8a0827b7344b2bee6da8eabef1d400e#r1567919332

examples/treeple/treeple_tutorial_pvalue_multiview.py
PSSF23: I don't think so. It could act as a conditional statement for users if they decide to modify the data.

sampan501 (approved):
Looks good to me, pending the larger changes in #257

PSSF23 (commented):
I'll incorporate the comments in a new PR. Feel free to comment under #257 as well if you have new ideas.

PR #258, opened 2024-04-17
FIX reverse simulation labels

PSSF23:
Resolving downstream problems after #254


PSSF23 (commented):
I forced the `sphinx` version to be 7.2.6 to avoid a useless warning in building doc.

PR #260, opened 2024-04-19
ENH optimize & add treeple tutorial figures

PSSF23:
- Reorder by single view and multiview
- Optimize figure style to be consistent with paper using: [1d](https://github.com/neurodata/might/blob/fig8_true_plots/Fig1.ipynb) & [2d](https://github.com/neurodata/might/blob/main/Comight_Fig1.ipynb)
- Add mutual information difference figure
- Select thumbnail for specific tutorials
- Output all statistics
- Add hellinger distance tutorial
- Add 2d plots for permuted 2-view data
- Change text color in the note box to blue

### DIFF
--- a/examples/treeple/treeple_tutorial_CMI.py
+++ b/examples/treeple/treeple_tutorial_CMI.py
@@ -1,7 +1,7 @@
 """
-====================================
-Treeple tutorial for calculating CMI
-====================================
+=====================
+2-1b: Calculating CMI
+=====================
 """
 
 import matplotlib.pyplot as plt
--- a/examples/treeple/treeple_tutorial_CMI.py
+++ b/examples/treeple/treeple_tutorial_CMI.py
@@ -6,6 +6,7 @@
 
 import matplotlib.pyplot as plt
 import numpy as np
+import pandas as pd
 import seaborn as sns
 from scipy.stats import entropy
 
@@ -17,6 +18,7 @@ from sktree.tree import MultiViewDecisionTreeClassifier
 sns.set(color_codes=True, style="white", context="talk", font_scale=1.5)
 PALETTE = sns.color_palette("Set1")
 sns.set_palette(PALETTE[1:5] + PALETTE[6:], n_colors=9)
+sns.set_style("white", {"axes.edgecolor": "#dddddd"})
 # %%
 # CMI
 # ---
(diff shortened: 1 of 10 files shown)


examples/treeple/treeple_tutorial_CMI.py
```diff
@@ -1,18 +1,24 @@
 """
-====================================
-Treeple tutorial for calculating CMI
-====================================
+=====================
+2-1b: Calculating CMI
```
adam2392, line 3: I don't think most ppl will know the order of our figures. Can we remove the numbering in the title? We can still order the sphinx examples as how it shows up tho.

examples/treeple/treeple_tutorial_CMI.py
```diff
@@ -1,18 +1,24 @@
...
+import pandas as pd
+import seaborn as sns
 from scipy.stats import entropy
 
 from sktree.datasets import make_trunk_classification
 from sktree.ensemble import HonestForestClassifier
 from sktree.stats import build_hyppo_oob_forest
 from sktree.tree import MultiViewDecisionTreeClassifier
 
+sns.set(color_codes=True, style="white", context="talk", font_scale=1.5)
```
adam2392, line 18: Maybe add a comment above stating we are using some formatting to make the figure look as close to "TODO" insert citation to the paper.

examples/treeple/treeple_tutorial_CMI.py
PSSF23: This is not related to the figures, just to separate single view and multiview. If we can order them with other methods, that would be great.

adam2392: https://sphinx-gallery.github.io/stable/gen_modules/sphinx_gallery.sorting.html

We can prolly just rename the filenames and apply a sorting based on that if we want the titles to be more specific.

E.g. `FileNameSortKey`

examples/treeple/treeple_tutorial_CMI.py
PSSF23: I don't feel that's necessary. It's just a figure style? Definitely agree on citing the paper somewhere when it's out.

PSSF23 (commented):
The images generated by sphinx has the axis labels cut off. Trying to resolve.

PR #264, opened 2024-04-26
FIX Raise error when missing-values encountered in scikit-tree trees

### LINKED ISSUE #263: Current status of missing values (in oblique trees)
This is a request for clarification.

The documentation for v0.7.X says that supervised trees (such as oblique decision trees) do not support missing values: https://github.com/neurodata/scikit-tree/blob/v0.7.0/doc/modules/supervised_tree.rst (jump to "Limitations compared to decision trees")

However, the following Python code runs without errors:
```python
from sklearn.datasets import load_iris
from sktree.tree import ObliqueDecisionTreeClassifier

import numpy

iris_X, iris_y = load_iris(return_X_y = True, as_frame = True)

# Make the feature matrix 25% sparse
iris_X = iris_X.mask(numpy.random.random(iris_X.shape) < .25)

classifier = ObliqueDecisionTreeClassifier()
classifier.fit(iris_X, iris_y)

pred_proba = classifier.predict_proba(iris_X)
print(pred_proba)
```

If missing values are really not supported, then I would expect the `ObliqueDecisionTreeClassifier.fit(X, y)` method to fail quickly and cleanly with an appropriate error message.

The `ObliqueTree.missing_go_to_left` attribute is definitely set. But I can see that its elements hold values that are not valid according to Scikit-Learn's missing-go-to-left conventions (ie. all elements should be either `0` or `1`, but the "active" values appear arbitrary 1-byte integers with values up to `128`).

I was experimenting with a setup where a "oblique projection feature" computed to missing value when any of its input features was a missing value, and then scoring the oblique tree using Scikit-Learn's missing-go-to-left algorithm. However, the predictions didn't seem to agree, which suggests that Scikit-Tree is doing something differently.

TLDR: As of today, is it permitted to pass missing values into oblique tree-based estimators or not?

adam2392:
Fixes #263 

- raises a ValueError when missing-value encountered in scikit-tree estimators

## Before submitting

## After submitting

### DIFF
--- a/sktree/tree/_neighbors.py
+++ b/sktree/tree/_neighbors.py
@@ -64,3 +64,7 @@ class SimMatrixMixin:
             The similarity matrix among the samples.
         """
         return compute_forest_similarity_matrix(self, X)
+
+    def _more_tags(self):
+        # XXX: no scikit-tree estimators support NaNs as of now
+        return {"allow_nan": False}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,6 +61,7 @@ jobs:
         run: |
           brew install ccache
           brew install gcc
+          brew install gettext
 
       - name: show-gcc
         run: |
--- a/build_requirements.txt
+++ b/build_requirements.txt
@@ -1,9 +1,9 @@
-meson
-meson-python
-cython>=3.0.8
+meson>=1.4.0
+meson-python>=0.16.0
+cython>=3.0.10
 ninja
 numpy
-scikit-learn>=1.4.1
+scikit-learn>=1.4.2
 click
 rich-click
 doit
--- a/doc/whats_new/v0.8.rst
+++ b/doc/whats_new/v0.8.rst
@@ -17,7 +17,7 @@ Changelog
     did not raise an error, and silently ran, assuming the missing-values were
     encoded as infinity value. This is now fixed, and the estimators will raise an
     ValueError if missing-values are encountered in ``X`` input array.
-    By `Adam Li`_ (:pr:`#263`)
+    By `Adam Li`_ (:pr:`#264`)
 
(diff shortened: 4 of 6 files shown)


sktree/tree/_neighbors.py
```diff
@@ -64,3 +64,7 @@ def compute_similarity_matrix(self, X):
             The similarity matrix among the samples.
         """
         return compute_forest_similarity_matrix(self, X)
+
+    def _more_tags(self):
+        # XXX: no scikit-tree estimators support NaNs as of now
```
vruusmann, line 69: Scikit-Tree appears to be based on Scikit-Learn 1.3.X, which supports missing values in standalone decision tree models (`DecisionTreeClassifier`, `DecisionTreeRegressor`) by default. The support for missing values in random forest models (`RandomForestClassifier` and `RandomForestRegressor`) was added in 1.4.X.

In that regard, perhaps the `_more_tags` method should be base class-dependent (one tag for `DecisionTree` and another one for `RandomForest`)?

adam2392: Scikit-tree will depend on each version of scikit-learn successively very tightly for now, so we'll consider the next release of v0.8 to be tied to 1.4.x. Is that okay?

vruusmann: Right now, when there is no missing value support, then it would be okay to attach the same "disable all"-type`_more_tags` method to the tree ABC, so that it affects all tree and tree ensemble estimators.

The exact SkTree <-> SkLearn version matching is not critical until SkTree really starts enabling missing values.

adam2392:
I will merge this if LGTYou @vruusmann?

adam2392:
CI failures seem unrelated: https://github.com/actions/setup-python/issues/855

Issue #257, opened 2024-04-17
ENH treeple tutorials should use native functions in scikit-tree


sampan501:
From #256:

We should fix the tutorials to use native functions in scikit-tree, like `build_coleman_forest` etc. It should be clear to the user that they don't need to write their own custom function to compute a p-value using MIGHT.

PSSF23:
Jovo wants to add a figure similar to[ this one](https://www.researchgate.net/figure/The-null-distribution-of-the-test-statistic-in-4-for-the-null-hypothesis-that-there-is_fig1_304858060), which would require all the intermediate statistics (the ones that 2 permuted forest make). `build_coleman_forest` doesn't return those values, unfortunately, so I'll postpone the conversion for now.

sampan501:
Sure, we could write it like this instead. Show all the intermediate stats figures, and at the end show how to compute the stat and p-value using treeple

adam2392:
The statistics we need are the null distribution? What else?

adam2392:
I assume this has been addressed in #280 and #273. Feel free to reopen if not.

PR #274, opened 2024-05-06
MAINT `HonestForestClassifier` now has `bootstrap=True` as the default. And removing old MIGHT code

### LINKED ISSUE #146: Should `HonestForest*` have `bootstrap=False` or `bootstrap=True` as default
It is unclear what the default should be because in scikit-learn, `bootstrap=True` on Forests are the default.

cc: @rflperry @sampan501 mentioned that your original implementation had `boostrap=False` as the default. To my knowledge, there is no reason to default in `HonestForests`, so I'm wondering if we should stick w/ scikit-learn defaults?

adam2392:
Fixes #146 

- bootstrap=True is the default now
- Removed all instances of `FeatureImportanceClassifier` and other MIGHT methods that are outdated. This is in favor of the functional API used for (CO)MIGHT
- There was a small bug in `build_coleman_forest` stemming from #281 and #257 where the PermutationHonestForest should get passed `covariate_index`. It is not tested in `test_comight_repeated_feature_sets`.

## Before submitting

## After submitting


sampan501 (approved):
LGTM!

sampan501:
Pending unit test pass

sktree/stats/tests/test_coleman.py
adam2392, line 1: This will get refactored to use the latest API

PR #281, opened 2024-06-06
MAINT Fix CI

adam2392:

- updates the CI to leverage new `spin` version v0.10+
- patch some unit-test failures that arise from https://github.com/neurodata/scikit-tree/pull/273/files#diff-f5b103dd1c3449d5467e6833a115f8b2bd1b074cc709ca3a548a696abf165a3e

Follows the pattern suggested in https://github.com/scientific-python/spin

- the current GH actions that generates the coverage report seems to be broken, but may be an issue with versioning and should get patched upstream in `coverage`, pytest and pytest-cov

## Before submitting

## After submitting


PSSF23 (commented):
As 3.9 and 3.10 are no longer supported in the Mac-latest tests, should we remove these two versions?

adam2392:
> As 3.9 and 3.10 are no longer supported in the Mac-latest tests, should we remove these two versions?

Can you clarify what you mean? I might've missed the context.

PSSF23:
@adam2392 It's in the conversation you sent before. They say that they no longer support lower python versions in Mac-latest meson tests.

PSSF23:
https://github.com/actions/setup-python/issues/855

sktree/stats/forestht.py
```diff
@@ -1307,6 +1307,10 @@ def build_coleman_forest(
         rng.shuffle(temp_col)
         X_null[:, covariate_index] = temp_col
 
+    if not isinstance(perm_est, PermutationHonestForestClassifier):
```
PSSF23, line 1299: This error needs to be removed because `PermutationHonestForestClassifier` is no longer used anywhere. And the tutorials use `HonestForestClassifier` directly with `build_coleman_forest`, so this error raising would cause problems.

adam2392: Besides simplicity, is there any reason to condense the two that I missed?

The reason the separate class is nice is for the future work, if we want to permute in more exotic ways. E.g. permute per tree, permute nearest-nbrs, etc. If the work all finishes, and we end up not needing that separate class, we can always revert to what we have here.

adam2392: Are lines 1299-1308 necessary if we use the separate class?

YuxinB (approved):

.github/workflows/main.yml
```diff
@@ -249,90 +253,68 @@ jobs:
...
+      #     conda update -y conda
+
+      #     # Add conda-forge channel and install openblas
+      #     conda config --add channels conda-forge
+      #     conda install -y openblas
+
+      - name: pip-packages
+        run: |
+          pip install spin
+          pip install .[build]
```
sampan501: why is this not also `python -m pip install -r build_requirements.txt` etc.

sampan501 (commented):
Just a question

.github/workflows/main.yml
adam2392: Good question. It should be!

adam2392: I fixed it now

sampan501 (approved):

PR #291, opened 2024-07-03
FIX ignore nan values when summing posteriors

### LINKED ISSUE #290: BUG predict_proba() not compatible with ignore setting for empty leaf
When using the "ignore" option for empty leaf, the sum of posterior estimates would be `nan` if even only one tree encounters an empty leaf. When the number of trees is high, this will result in all `nan`'s for `predict_proba()`

https://github.com/neurodata/scikit-tree/blob/98b67fcbaf028ee8a9bd92407089ab4a34008e9b/sktree/ensemble/_honest_forest.py#L824-L829

PSSF23:
Close #290 

- Fix `np.nan` posteriors when `honest_prior="ignore"`
- Set "ignore" as default prior
- Add & optimize unit tests


PSSF23 (commented):
Seems a simple and straightforward fix to me. I'll write unit tests about it, but first want to see if it cause other problems.

PSSF23 (commented):
This test is not checking the ignore setting properly: `test_impute_posteriors`. As it doesn't give an error when there's too many `nan`'s.

The fix is also incomplete, as `np.nansum` gives zeros when all the tree posteriors are `nan`'s. `impute_missing` is not working as it should. I'm forcing it to work on `zero mask posteriors`, as the other two leaf correction methods would not result in zeros.

sktree/tests/test_honest_forest.py
```diff
@@ -263,6 +263,27 @@ def test_impute_posteriors(honest_prior, val):
         ), f"Failed with {honest_prior}, prior {clf.estimators_[0].empirical_prior_}"
 
 
+@pytest.mark.parametrize(
+    "honest_prior, val",
+    [
+        ("ignore", np.nan),
+    ],
+)
```
adam2392: This parametrization doesn't seem necessary

sktree/tests/test_honest_forest.py
```diff
@@ -263,6 +263,27 @@ def test_impute_posteriors(honest_prior, val):
...
+def test_ignore_posteriors(honest_prior, val):
```
adam2392:
```suggestion
def test_honestforest_predict_proba_with_honest_prior(honest_prior, val):
```

This test seems like it should conceptually test `predict_proba` for the HonestForestClassifier for different honest_priors?

sktree/tests/test_honest_forest.py
```diff
@@ -263,6 +263,27 @@ def test_impute_posteriors(honest_prior, val):
...
+        honest_fraction=0.5, random_state=0, honest_prior=honest_prior, n_estimators=100
+    )
+    clf = clf.fit(X, y)
+
+    y_proba = clf.predict_proba(X)
+
+    assert (
+        len(np.where(np.isnan(y_proba[:, 0]))[0]) < 10
+    ), f"Failed with {honest_prior}, prior {clf.estimators_[0].empirical_prior_}"
+
```
adam2392: Perhaps it should just assert that none of `y_pred` are nans? I think with enough trees this is possible. 

It would be nice to add an in-line comment specifying this is a requirement for the test to pass too.

sktree/ensemble/_honest_forest.py
```diff
@@ -672,10 +672,7 @@ def _predict_proba(self, X, indices=None, impute_missing=None):
         zero_mask = posteriors.sum(2) == 0
         posteriors[~zero_mask] /= posteriors[~zero_mask].sum(1, keepdims=True)
 
-        if impute_missing is None:
-            pass
-        else:
-            posteriors[zero_mask] = impute_missing
+        posteriors[zero_mask] = impute_missing
```
adam2392: This implies that some of `posteriors` array is `None`, which would be bad as I don't think this is equivalent to `np.nan`. I'm surprised we don't run into any unit-test failures, but I guess this is a private function.

Instead, I would say: `if impute_missing is not None: posteriors[zero_mask] = impute_missing`.

adam2392 (changes requested):
Left a few comments.

Re CI failures:

```
FAILED sktree/stats/tests/test_forestht.py::test_small_dataset_dependent[0] - AssertionError
FAILED sktree/tests/test_extensions.py::test_predict_proba_per_tree[HonestForestClassifier-2] - AssertionError
FAILED sktree/tests/test_extensions.py::test_predict_proba_per_tree[HonestForestClassifier-3] - AssertionError
```

I would recommend:

1. test_small_dataset_dependent, just adding the `honest_prior='empirical'` to make the test backwards compatible.
2. test_predict_proba_per_tree, add an extra kwarg for `HonestForestClassifier` for the `honest_prior='empirical'`. You can also add a test for `test_predict_proba_per_tree` for HonestForestClassifier if `honest_prior='ignore'`.

PSSF23 (commented):
@adam2392 you think this is mergeable?

doc/whats_new/v0.9.rst
```diff
@@ -21,7 +21,10 @@ Changelog
 ---------
 
 - |API| Rename the package to ``treeple``. By `SUKI-O`_ (:pr:`#292`)
-
+- |Fix| Fixed a bug in the :class:`treeple.HonestForestClassifier` where posteriors
```
adam2392:
```suggestion
- |Fix| Fixed a bug in the predict_proba function of the :class:`treeple.HonestForestClassifier` where posteriors
```

adam2392 (approved):

adam2392:
Thanks for the PR @PSSF23

Issue #20, opened 2023-02-03
Implementation of general MORF trees (Gabor, Fourier, Gaussian, and generic kernels)


adam2392:
## Basic MORF implementation

Need to implement:

1. Cython splitter: Patch splitter (discontiguous and contiguous)
2. Cython tree: I think we don't even need a new Cython tree. We can just use the ObliqueTree class
3. Python interface: Implement a new Python API for morf

The main difference in the MORF splitter is that we need to take into account the shapes of the input dataset. The hardest thing to implement is the splitter. We just need to replicate the splitter in the MORF paper (weights of 1). Once we have that, the next set of goals will be to implement a more exotic splitter, where the weights and the shape of the patch is some kernel function (e.g. Gabor).

## Advanced MORF implementation

I think the first thing we should try is the Gabor-tree. To my knowledge there is no robust C++/C/Cython implementation of Gabor filter generation. Therefore, what we can do is generate a "filter-bank" in Python using scikit-image. Then we pass this filter-bank from Python via read-only memoryviews into Cython.

The open technical coding question is how to do this efficiently. You cannot pass a list of ndarrays as this is not supported in Cython. There are a bunch of possible solns. floated here https://stackoverflow.com/questions/50439908/array-of-memoryviews-in-cython.

Ref: https://scikit-image.org/docs/stable/auto_examples/features_detection/plot_gabor.html

Previous discussion: https://github.com/NeuroDataDesign/manifold_random_forests/issues/3

## Misc thoughts
Rather than apply random projections using a kernel, we could also take each randomly chosen kernel and apply an entire convolution on the data. (https://github.com/dagss/cython-tutorial/blob/master/examples/convolve1/convolve.py). This would result in a new representation of the data for each kernel, and could be used to determine the "best kernels". One might for example take the resulting kernel convolution outputs and sort them based on which one produces the best criterion metric (i.e. differentiation among the classes). This is more expensive because the criterion needs to loop over a bigger vector, but may be worth it for the sake of choosing the "best kernels".

Another approach for making the implementation fast/robust enough is to somehow create a dependency on openCV and leverage their C++ code.

adam2392:
@jovo the pseudocode I am thinking of to implement a Gabor tree is follows. Some feedback and suggestions of who might be able to assist on technical details would be helpful

1. define grid/range of parameters for a certain filter (i.e. Gabor)
2. sample these filters to construct our filter bank
3. at each node compute split candidates by:
[...]

PR #302, opened 2024-07-23
FEA Kernel decision trees with user-defined kernel dictionaries

adam2392:
#### Reference Issues/PRs
Towards: #20 
Closes: https://github.com/neurodata/treeple/pull/269

- Adds kernel decision trees


adam2392:
Notes:

1. fused-type util functions to allow vector[intp_t] and intp_t[:] is useful and should be merged upstream
2. the performance penalty of MORF most likely is in space and in time. We iterate over the patches, which may become non-trivial size multiples times. The first time is in generating the patch vectorized indices and storing these in the projection matrix. The second time is in actually applying the patch vectorized indices to obtain the feature values. First of all, we shouldn't need to store the entire projection matrix, as this may potentially scale up the space complexity. Second of all, we should only need to iterate once. 

To fix 2., we need to decouple the assumptions of the design with that of the oblique splitter, where storing all this stuff and iterating more than once has negligible cost.

Instead for MORF, we may want to instantiate a separate subclass that isn't an extension of oblique. Here, we want to store the "parameters" of the projection matrix per split since we need some way of indexing what was the "best split" and then getting all the relevant information needed in order to apply it. The current way of doing it leverages the fact that one can simultaneously iterate over projection_weights and projection_indices to obtain the feature value during `predict`. So we should maybe piggyback that method, but we only store the final projection_indices and weights once we've selected the "best split".

To parameterize a split indices, we need the top left point and the corresponding patch dims. To parameters a split weight, we need the relevant kernel. In the naive case, this is just all values of "1". In the case of a gaussian kernel for instance, The # of non-zeros would correspond to the number of split indices. So we can either i) store the parameters of the gaussian kernel (mu, sigma, shape), ii) the full set of weights, or iii) the index of kernel in the pre-generated kernel dictionary (passed in by the user). Obviously we don't want to store the full sets of weights, so that's not viable. So i) and iii) seem most viable, where i) requires then one to write a function to generate the full kernel given the parameters, and iii) just requires keeping a reference to the original kernel dictionary generated by the user.

PR #311, opened 2024-08-08
dropped bottleneck warning message

### LINKED ISSUE #310: Bottleneck should not raise a warning unless relevant function in treeple.stats is called
These lines seem like they are ran regardless of whether or not treeple.stats is used, since they are in the global namespace.

```
if BOTTLENECK_AVAILABLE and DISABLE_BN_ENV_VAR not in os.environ:
    nanmean_f = bn.nanmean
    anynan_f = lambda arr: bn.anynan(arr, axis=2)
else:
    warnings.warn(
        "Not using bottleneck for calculations involvings nans. Expect slower performance."
    )
    nanmean_f = np.nanmean
    anynan_f = lambda arr: np.isnan(arr).any(axis=2)
```

so I get this warning consistently no matter what

```
/Users/adam2392/Documents/treeple/treeple/stats/utils.py:34: UserWarning: Not using bottleneck for calculations involvings nans. Expect slower performance.
  warnings.warn(
```

@ryanhausen do you mind submitting a PR to patch this behavior?

ryanhausen:
#### Reference Issues/PRs

Fixes #310

Drops a bottleneck warning that appears every import of `treeple.stats`, which can be too much if the user isn't using the functions that depend on bottleneck for nans.

#### Any other comments?

Rather than move the warning, it makes sense to drop it for now. I am working on a  sparse implementation of some of the `treeple.stats` calculations that will involve a refactor and so the warning may have a better home somewhere else when I am done.

### DIFF
--- a/treeple/stats/utils.py
+++ b/treeple/stats/utils.py
@@ -1,6 +1,5 @@
 import os
 import sys
-import warnings
 from typing import Optional, Tuple
 
 import numpy as np
--- a/treeple/stats/utils.py
+++ b/treeple/stats/utils.py
@@ -1,5 +1,6 @@
+import importlib.util
 import os
-import sys
+import warnings
 from typing import Optional, Tuple
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


adam2392:
We can wait to see the sparse refactor then. I think a warning message is still warranted when the user uses the relevant functions (not during import)

ryanhausen:
Closing this out. #317 has this incorporated into it.

PR #317, opened 2024-08-20
Sparse Implementation of `build_coleman_forest` plus a fix for bottleneck import

### LINKED ISSUE #310: Bottleneck should not raise a warning unless relevant function in treeple.stats is called

ryanhausen:
#### Reference Issues/PRs

Fixes #310 

This adds a sparse implementation of `build_coleman_forest` to `treeple.stats`. The sparse implementation relies on `scipy.stats.sparse` and so only works in the case of binary classification and regression. In my tests using `cProfile`/`memray`, the sparse implementation is a little over 50% faster and uses about %7 less memory, see [pdf](https://github.com/user-attachments/files/16681054/sparse_profiling.pdf) for more information. I am also attaching the [profiling data](https://github.com/user-attachments/files/16681063/sparse_profiling.zip)
 if you'd like to take a look.

This PR also includres a change to the bottleneck implementation to for the dense implementation of `build_coleman_forest`. Specifically there are two important changes:
1. The check for the existence of the bottleneck package is now done using `importlib.util.find_spec("bottleneck")`. The old check using `sys` seemed to work fine and passed tests, but actually didn't seem work unless bottleneck had been imported before, which when the tests are run is true, but may not be the case in other situations
2. the aliased and lambda functions `nanmean_f` and `anynan_f` are now defined using `def` so that the warning message can moved into their respective calls rather than at import.

#### Any other comments?

The scipy.stats sparse implementation should be swapped out for pydata.sparse when that implementation is performant enough, as it is more general.

### DIFF
--- a/treeple/stats/forest.py
+++ b/treeple/stats/forest.py
@@ -107,7 +107,8 @@ def build_coleman_forest(
     return_posteriors : bool, optional
         Whether or not to return the posteriors, by default True.
     use_sparse : bool, optional
-        Whether or not to use a sparse representation for the posteriors.
+        Whether or not to use a sparse for the calculation of the permutation
+        statistics, by default False. Doesn't affect return values.
     **metric_kwargs : dict, optional
         Additional keyword arguments to pass to the metric function.
 
@@ -173,11 +174,23 @@ def build_coleman_forest(
         # sparse indices and values with an array
         if return_posteriors:
             n_trees = y_pred_proba_orig_perm.shape[0] // 2
+            n_samples = y_pred_proba_orig_perm.shape[1]
             # slicing a csc matrix this way is not efficient, but it is
             # it is only done once, so I am not sure if it is worth it to
             # optimize this.
-            orig_forest_proba = y_pred_proba_orig_perm[:n_trees, :]
-            perm_forest_proba = y_pred_proba_orig_perm[n_trees:, :]
+            to_coords_data = lambda x: (x.row.astype(int), x.col.astype(int), x.data)
+
+            row, col, data = to_coords_data(y_pred_proba_orig_perm[:n_trees, :].tocoo())
+            orig_forest_proba = np.full((n_trees, n_samples), np.nan, dtype=np.float64)
+            orig_forest_proba[row, col] = data
+
+            row, col, data = to_coords_data(y_pred_proba_orig_perm[n_trees:, :].tocoo())
+            perm_forest_proba = np.full((n_trees, n_samples), np.nan, dtype=np.float64)
+            perm_forest_proba[row, col] = data
+
+            if y.shape[1] == 2:
+                orig_forest_proba = np.column_stack((orig_forest_proba, 1 - orig_forest_proba))
+                perm_forest_proba = np.column_stack((perm_forest_proba, 1 - perm_forest_proba))
     else:
         metric_star, metric_star_pi = _compute_null_distribution_coleman(
             y,
--- a/treeple/stats/forest.py
+++ b/treeple/stats/forest.py
@@ -12,7 +12,12 @@ from sklearn.utils.multiclass import type_of_target
 from .._lib.sklearn.ensemble._forest import ForestClassifier
 from ..tree._classes import DTYPE
 from .permuteforest import PermutationHonestForestClassifier
-from .utils import METRIC_FUNCTIONS, POSITIVE_METRICS, _compute_null_distribution_coleman
+from .utils import (
+    METRIC_FUNCTIONS,
+    POSITIVE_METRICS,
+    _compute_null_distribution_coleman,
+    _compute_null_distribution_coleman_sparse,
+)
 
 
 def _parallel_predict_proba_oob(predict_proba, X, out, idx, test_idx, lock):
@@ -29,6 +34,20 @@ def _parallel_predict_proba_oob(predict_proba, X, out, idx, test_idx, lock):
     with lock:
         out[idx, test_idx, :] = prediction[test_idx, :]
     return prediction
+
+
+def _parallel_predict_proba_oob_sparse(predict_proba, X, test_idx):
+    """
+    This is a utility function for joblib's Parallel.
+    It can't go locally in ForestClassifier or ForestRegressor, because joblib
+    complains that it cannot pickle it when placed there. Different from
+    _parallel_predict_proba_oob, this function returns the predictions and
+    indices for those values that are finite (not nan or inf).
+    """
+    # each tree predicts proba with a list of output (n_samples, n_classes[i])
+    prediction = predict_proba(X[test_idx, :], check_input=False)
+    good_value_mask = np.isfinite(prediction[:, 0])
+    return test_idx[good_value_mask], prediction[good_value_mask]
 
 
 ForestTestResult = namedtuple(
@@ -48,6 +67,7 @@ def build_coleman_forest(
     verbose=False,
     seed=None,
     return_posteriors=True,
+    use_sparse=False,
     **metric_kwargs,
 ):
     """Build a hypothesis testing forest using a two-forest approach.
@@ -86,6 +106,8 @@ def build_coleman_forest(
         Random seed, by default None.
     return_posteriors : bool, optional
         Whether or not to return the posteriors, by default True.
+    use_sparse : bool, optional
+        Whether or not to use a sparse representation for the posteriors.
     **metric_kwargs : dict, optional
         Additional keyword arguments to pass to the metric function.
 
@@ -112,14 +134,19 @@ def build_coleman_forest(
     metric_func: Callable[[ArrayLike, ArrayLike], float] = METRIC_FUNCTIONS[metric]
 
     # build two sets of forests
-    est, orig_forest_proba = build_oob_forest(est, X, y, verbose=verbose)
+    est, orig_forest_proba = build_oob_forest(est, X, y, use_sparse=use_sparse, verbose=verbose)
 
     if not isinstance(perm_est, PermutationHonestForestClassifier):
         raise RuntimeError(
             f"Permutation forest must be a PermutationHonestForestClassifier, got {type(perm_est)}"
         )
     perm_est, perm_forest_proba = build_oob_forest(
-        perm_est, X, y, verbose=verbose, covariate_index=covariate_index
+        perm_est,
+        X,
+        y,
+        use_sparse=use_sparse,
+        verbose=verbose,
+        covariate_index=covariate_index,
     )
 
     # get the number of jobs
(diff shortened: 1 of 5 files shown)


treeple/stats/forest.py
```diff
@@ -86,6 +106,8 @@ def build_coleman_forest(
         Random seed, by default None.
     return_posteriors : bool, optional
         Whether or not to return the posteriors, by default True.
+    use_sparse : bool, optional
+        Whether or not to use a sparse representation for the posteriors.
```
adam2392: Will the returned predictions and posteriors also be "sparse"? If so, perhaps we can document this. I assume structurally they are the same and can be converted via `todense` and `to_csc`?

treeple/stats/utils.py
```diff
@@ -20,22 +21,34 @@
...
 else:
-    warnings.warn(
-        "Not using bottleneck for calculations involvings nans. Expect slower performance."
-    )
-    nanmean_f = np.nanmean
-    anynan_f = lambda arr: np.isnan(arr).any(axis=2)
+
+    def nanmean_f(arr: ArrayLike, axis=None) -> ArrayLike:
+        warnings.warn(BOTTLENECK_WARNING)
+        return np.nanmean(arr, axis=axis)
```
adam2392, line 47: It seems this is not tested? Perhaps we can use the environment variable setting to ensure the warning gets triggered when `bottleneck` is not used?

treeple/stats/utils.py
```diff
@@ -316,3 +332,194 @@ def get_per_tree_oob_samples(est: BaseForest):
...
         oob_samples.append(unsampled_indices)
     return oob_samples
+
+
+def _get_forest_preds_sparse(
+    all_y_pred: sp.csc_matrix,  # (n_trees, n_samples)
+    all_y_indicator: sp.csc_matrix,  # (n_trees, n_samples)
+    forest_indices: ArrayLike,  # (n_trees,)
+) -> ArrayLike:
+    """Get the forest predictions for a set of trees using sparse matrices."""
```
adam2392: Can you document the input and output in a docstring?

treeple/stats/utils.py
```diff
@@ -316,3 +332,194 @@ def get_per_tree_oob_samples(est: BaseForest):
...
+    n_outputs = y_pred_proba_normal[1][0].shape[1]
+
+    all_oob_idxs = y_pred_proba_normal[0] + y_pred_proba_perm[0]
+    tree_row_ids = np.repeat(
+        np.arange(n_trees, dtype=np.int32),
+        [len(oob) for oob in all_oob_idxs],
+    )  # (n_oob_samples)
+    all_oob_idxs = np.concatenate(all_oob_idxs, axis=0)  # (n_oob_samples)
+
+    # Because we are assuming 2D sparse matrices for binary classification or
```
adam2392:
```suggestion
    # XXX: Because we are assuming 2D sparse matrices for binary classification or
```

treeple/stats/utils.py
```diff
@@ -316,3 +332,194 @@ def get_per_tree_oob_samples(est: BaseForest):
...
+
+    oob_predictions = sp.coo_matrix(
+        (all_oob_values, (tree_row_ids, all_oob_idxs)),
+        shape=(n_trees, len(y_test)),
+    ).tocsc()  # (n_trees, n_samples)
+
+    oob_indicators = sp.coo_matrix(
+        (np.ones_like(all_oob_values, dtype=np.uint8), (tree_row_ids, all_oob_idxs)),
+        shape=(n_trees, len(y_test)),
+    ).tocsc()  # (n_trees, n_samples)
```
adam2392: Do we first construct `coo` and then convert to `csc` for any particular reason? Perhaps good to state it in an in-line comment.

adam2392 (commented):
Cool! Thanks for the profiling data and documentation. It's very convincing. 

Left a few comments to tighten up the PR

treeple/stats/forest.py
ryanhausen: Ah good catch this should be documented. Currently they are sparse csc matrices, but actually that behavior will be different than the dense behavior because the dense matrix has nans and the sparse doesn't. I'll change it to return dense matrices that are the same as the dense implementation.

treeple/stats/utils.py
ryanhausen: good catch, the csc matrix can be instantiated using `coo` format, so I'll simplify those constructors

treeple/stats/utils.py
ryanhausen: This is weird. For some reason when I run this through pytest this code doesn't seem to be acknowledged, even `print` statements don't get run. If I run the function manually e.g. something like:

```
if __name__ == "__main__":
    test_build_coleman_forest(False, False)
```

then the behavior seems to be correct. I am going to work on it. I may just split the test out. I'll follow up when I have a fix.

ryanhausen: Ok they should be covered now. I tested it separately.

ryanhausen:
@adam2392 I have updated the PR, let me know if I missed anything or if you want to see some more changes.

treeple/stats/forest.py
```diff
@@ -112,36 +135,78 @@ def build_coleman_forest(
...
+            # optimize this.
+            to_coords_data = lambda x: (x.row.astype(int), x.col.astype(int), x.data)
+
+            row, col, data = to_coords_data(y_pred_proba_orig_perm[:n_trees, :].tocoo())
+            orig_forest_proba = np.full((n_trees, n_samples), np.nan, dtype=np.float64)
+            orig_forest_proba[row, col] = data
+
+            row, col, data = to_coords_data(y_pred_proba_orig_perm[n_trees:, :].tocoo())
+            perm_forest_proba = np.full((n_trees, n_samples), np.nan, dtype=np.float64)
+            perm_forest_proba[row, col] = data
```
adam2392, line 188: Can you explain what's going on here? I guess it is to make the return in "dense" format?

treeple/stats/tests/test_forest.py
```diff
@@ -235,20 +236,22 @@ def test_comight_repeated_feature_sets(seed):
...
-    else:
+    if not use_bottleneck and utils.DISABLE_BN_ENV_VAR not in os.environ:
         os.environ[utils.DISABLE_BN_ENV_VAR] = "1"
-        importlib.reload(utils)
-        importlib.reload(stats)
+    elif use_bottleneck and utils.DISABLE_BN_ENV_VAR in os.environ:
+        del os.environ[utils.DISABLE_BN_ENV_VAR]
+
+    importlib.reload(utils)
+    importlib.reload(stats)
```
adam2392, line 256:
```suggestion
    # to allow testing without bottleneck
    importlib.reload(utils)
    importlib.reload(stats)
```

Is this accurate?

adam2392 (commented):
Mostly lgtm. Just some small nits

treeple/stats/tests/test_forest.py
ryanhausen: yep, I added a comment

treeple/stats/forest.py
ryanhausen: Yes, I added a comment. Your understanding is correct. Let me know if you would like more clarification.

ryanhausen:
@adam2392 I added some documentation to address your comments. Let me know what you think.

adam2392 (approved):

adam2392:
I will let @sampan501 @SUKI-O @YuxinB @PSSF23 take a look as well.

treeple/stats/utils.py
```diff
@@ -316,3 +329,209 @@ def get_per_tree_oob_samples(est: BaseForest):
         )
         oob_samples.append(unsampled_indices)
     return oob_samples
+
+
+def _get_forest_preds_sparse(
```
sampan501, line 334: maybe I missed it somewhere, but are there unit tests for these methods anywhere?

adam2392: @ryanhausen

ryanhausen: @sampan501, @adam2392  No there isn't a test for this function in test_utils.py. This function, `_get_forest_preds_sparse`, is a good candidate for a unit test though. I can write one up. 

I worry the other functions: `_parallel_build_null_forests_sparse` and `_compute_null_distribution_coleman_sparse`, like their dense counterparts, do too much to be easily unit tested and are tested via the integration tests. However,  if you would like, we can refactor the dense/sparse functions into smaller units so that they can all have tests written. Would probably be cleaner, but would take time. 

What would you like to see?

adam2392: For now, I think it's fine the way it is. If we find means of re-designing the internal API to make things simpler and easier to test, then it's warranted, but I don't see a clear path. Do you?

ryanhausen: I think it would take some thought. There is some repetition we might be able to leverage, but I am not sure if it would come at the cost of clarity. So just write up a test for `_get_forest_preds_sparse`?

adam2392: Yeah I think that's sufficient then for now.

sampan501: sounds good. Maybe we can make an issue for having such unit tests/re-designing the API so we don't forget. Otherwise, I think this PR lgtm

ryanhausen:
@sampan501  @adam2392  Hey I added a test, but it looks like I some other tests are failing. I submitted an issue, #328. I'll start looking into it, but it might go faster if you have an intuition on what the issue is.

adam2392:
This is gonna be an issue due to parameter validation in scikit-learn. I will push a fix when I have time. It seems the rest of the code works tho.

I would prioritize other issues until I have the time to push up the fix

ryanhausen:
@adam2392 sounds good. Should we let this sit and update and merge after you have a fix for the other issue or merge as is? @sampan501

sampan501:
I'm happy to merge if @adam2392 is cool with it

sampan501 (approved):

adam2392:
> @adam2392 sounds good. Should we let this sit and update and merge after you have a fix for the other issue or merge as is? @sampan501

I will merge once the fix is added and the CI is confirmed to be happy. No rush on this one.

adam2392:
@ryanhausen if you update wrt `main`, the CIs should be fixed

adam2392:
Thanks @ryanhausen !

PR #323, opened 2024-09-09
FEA Add warning to control against runtime might when wanting to run comight

adam2392:

### DIFF
--- a/treeple/stats/forest.py
+++ b/treeple/stats/forest.py
@@ -128,7 +128,8 @@ def build_coleman_forest(
     if covariate_index is None and isinstance(est.tree_estimator, MultiViewDecisionTreeClassifier):
         warn(
             "Covariate index is not defined, but a MultiViewDecisionTreeClassifier is used. "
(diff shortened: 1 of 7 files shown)


treeple/stats/forest.py
```diff
@@ -111,13 +114,23 @@ def build_coleman_forest(
...
 
     if not isinstance(perm_est, PermutationHonestForestClassifier):
         raise RuntimeError(
             f"Permutation forest must be a PermutationHonestForestClassifier, got {type(perm_est)}"
         )
+
+    if covariate_index is None and isinstance(est.tree_estimator, MultiViewDecisionTreeClassifier):
+        warn(
+            "Covariate index is not defined, but a MultiViewDecisionTreeClassifier is used. "
+            "If using CoMIGHT, one should define the covariate index to permute."
```
sampan501: I think we need a sentence here saying that MIGHT is being run

sampan501 (changes requested):
1 minor request, otherwise lgtm!

PR #335, opened 2024-11-11
[ENH] sklearn 1.6.dev0 adjustments.

goraj:
Hi all,

New to the project but I've been loosely following @adam2392 and the project for a while now. 
I setup a dev environment according to DEVELOPMENT.md and ran into a few issues due to sklearn 1.6.dev0 being installed. Namely the introduction of ```check_sample_weight_equivalence``` in https://github.com/scikit-learn/scikit-learn/commit/364cafe979592abdf90b56722e18c27f3b02b9b2 leads to expected but not skipped test-case failures.

Changes will skip  check_sample_weight_equivalence testing for forest implementations. It also addresses some un-pickling issues encountered during testing due to joblib/loky in the structure of treeple.stats.utils.
Changes should be backward compatible.

### DIFF
--- a/treeple/stats/utils.py
+++ b/treeple/stats/utils.py
@@ -5,6 +5,7 @@ from typing import Optional, Tuple
 
 import numpy as np
 import scipy.sparse as sp
+from joblib import parallel_config
 from joblib import Parallel, delayed
 from numpy.typing import ArrayLike
 from scipy.stats import entropy
@@ -234,18 +235,20 @@ def _compute_null_distribution_coleman(
 
     # generate the random seeds for the parallel jobs
     ss = np.random.SeedSequence(seed)
-    out = Parallel(n_jobs=n_jobs)(
-        delayed(_parallel_build_null_forests)(
-            y_pred_ind_arr,
-            n_estimators,
-            all_y_pred,
-            y_test,
-            seed,
-            metric,
-            **metric_kwargs,
+    with parallel_config("multiprocessing"):
+        out = Parallel(n_jobs=n_jobs)(
+            delayed(_parallel_build_null_forests)(
+                y_pred_ind_arr,
+                n_estimators,
+                all_y_pred,
+                y_test,
+                seed,
+                metric,
+                **metric_kwargs,
+            )
+            for i, seed in zip(range(n_repeats), ss.spawn(n_repeats))
         )
-        for i, seed in zip(range(n_repeats), ss.spawn(n_repeats))
-    )
+
 
     for idx, (first_half_metric, second_half_metric) in enumerate(out):
         metric_star[idx] = first_half_metric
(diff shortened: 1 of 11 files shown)


goraj:
Nothing?

treeple/stats/utils.py
```diff
@@ -234,18 +234,19 @@ def _compute_null_distribution_coleman(
...
+            delayed(_parallel_build_null_forests)(
+                y_pred_ind_arr,
+                n_estimators,
+                all_y_pred,
+                y_test,
+                seed,
+                metric,
+                **metric_kwargs,
+            )
+            for i, seed in zip(range(n_repeats), ss.spawn(n_repeats))
```
adam2392, line 248: Why was this change made?

treeple/stats/utils.py
```diff
@@ -512,20 +513,21 @@ def _compute_null_distribution_coleman_sparse(
...
-            oob_predictions,
-            oob_indicators,
-            y_test,
-            n_outputs,
-            seed,
-            True,
-            metric,
-            **metric_kwargs,
+    with parallel_config("multiprocessing"):
+        out = Parallel(n_jobs=n_jobs)(
```
adam2392, line 517: Same here

adam2392 (changes requested):
Sorry for the delay @goraj and thanks for the PR!

I left a review. So I was thinking, should we just revamp how we're using the `parametrize_with_checks` function. Instead of labeling `sample_weight_equivalence` as an ignored test per area, perhaps let's introduce a `test_common.py` function under `treeple/tests/`, and then we can do the "sklearn compatible" check there. Then, we can consolidate what tests we want to ignore in the same file.

Kind of like here:

https://github.com/scikit-learn/scikit-learn/blob/66270e46b77d6202559bae4929ec83ab320beb1e/sklearn/utils/_test_common/instance_generator.py#L771

and how `expected_failed_checks` kwarg of `parametrize_with_checks` is used to ignore tests inside https://github.com/scikit-learn/scikit-learn/blob/66270e46b77d6202559bae4929ec83ab320beb1e/sklearn/tests/test_common.py#L118

WDYT?

goraj:
> Sorry for the delay @goraj and thanks for the PR!
> 
> I left a review. So I was thinking, should we just revamp how we're using the `parametrize_with_checks` function. Instead of labeling `sample_weight_equivalence` as an ignored test per area, perhaps let's introduce a `test_common.py` function under `treeple/tests/`, and then we can do the "sklearn compatible" check there. Then, we can consolidate what tests we want to ignore in the same file.
> 
> Kind of like here:
> 
> https://github.com/scikit-learn/scikit-learn/blob/66270e46b77d6202559bae4929ec83ab320beb1e/sklearn/utils/_test_common/instance_generator.py#L771
> 
> and how `expected_failed_checks` kwarg of `parametrize_with_checks` is used to ignore tests inside https://github.com/scikit-learn/scikit-learn/blob/66270e46b77d6202559bae4929ec83ab320beb1e/sklearn/tests/test_common.py#L118
> 
> WDYT?

Thank you.
I agree that would help quite a bit. I will update the PR accordingly, just a bit busy right now.

treeple/stats/utils.py
goraj: If I remember correctly the default loky would segfault during unit testing the *Oblique trees.

treeple/stats/utils.py
goraj: Same as above.

PR #339, opened 2024-12-12
FIX remove regressor multi_label tag

PSSF23:

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         os:  # see scipy wheels.yml GH actions
           - [ubuntu-22.04, manylinux, x86_64]
-          - [macos-12, macosx, x86_64]
+          - [macos-latest, macosx, x86_64]
           # - [macos-12, macosx_*, arm64]
           - [windows-2019, win, AMD64]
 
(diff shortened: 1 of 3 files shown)


treeple/tree/_classes.py
```diff
@@ -2441,7 +2441,7 @@ def __sklearn_tags__(self):
         # XXX: nans should be supportable in SPORF by just using RF-like splits on missing values
         # However, for MORF it is not supported
         tags = super().__sklearn_tags__()
-        tags.regressor_tags.multi_label = True
+        # tags.regressor_tags.multi_label = True
```
adam2392: Perhaps let's just add an inline comment saying "TODO: add regression support"?

adam2392 (approved):
LGTM

adam2392 (changes requested):
I think tho branch should get updated wrt main

.github/workflows/build_wheels.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       matrix:
         os:  # see scipy wheels.yml GH actions
           - [ubuntu-22.04, manylinux, x86_64]
-          - [macos-12, macosx, x86_64]
+          - [macos-latest, macosx, x86_64]
```
adam2392, line 32: For my education, why do we need this change?

PSSF23: Sorry for late reply. It's due to the drop of support for the previous Mac version.

PSSF23:
Merged for new changes on honest prune/kernel method #345

PR #345, opened 2025-02-07
ENH add kernel method parameter on posterior predictions

YuxinB:
(just outpit the leaves value)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ requires = [
   "setuptools<=65.5",
   "packaging",
   "Cython>=3.0.10",
-  "scikit-learn>=1.5.0",
+  "scikit-learn>=1.6.0",
   "scipy>=1.5.0",
   "numpy>=1.25; python_version>='3.9'"
 ]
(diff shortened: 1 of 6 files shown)


PSSF23 (commented):
@YuxinB As Adam suggested before, can you add a hyperparameter to turn this function on/off? The kernel method should be the default, but I believe we still need to preserve the standard averaging method.

YuxinB:
https://github.com/neurodata/treeple/commits/main/ can some one help me check commits here

YuxinB:
kay, will change later this week!

PSSF23 (changes requested):
@YuxinB You can ignore the check errors for the `sklearn` fork. But can you fix the style and lint errors? Thanks! After that I'll merge the changes.

PSSF23 (commented):
@YuxinB On second thought, we should not merge the changes until the parameter is implemented. We need to also make sure the forest level `predict_proba` outputs correctly when testing on an held-out set.

YuxinB:
okay, I will implement the parameter by the end of this week.

PSSF23 (changes requested):
Modify all the default values to what we agreed on.

PSSF23 (approved):
Merging to facilitate productivity.

PR #347, opened 2025-02-19
FIX correct the invoke of oblique trees by feature combinations

mamamum:
#### Reference Issues/PRs
args feature_combinations in multiview tree is not in use

Now feature_combinations can be correctly  given to feature_combinations_, if input feature_combinations is None, it will be defaultly set 1, which means not using oblique method


PSSF23 (changes requested):
You can put default hyperparameter in the init function instead of setting it specifically. Just make it `feature_combinatinos=1` instead of `None`.

PR #348, opened 2025-02-21
FIX add attribute for kernel method

PSSF23:


PSSF23 (commented):
A quick fix for the kernel method attribute.

PR #355, opened 2025-03-30
ENH enable multiview-oblique splitting & unlock multiclass restrictions

YuxinB:

### DIFF
--- a/treeple/tree/_oblique_splitter.pyx
+++ b/treeple/tree/_oblique_splitter.pyx
@@ -820,7 +820,7 @@ cdef class MultiViewSplitter(BestObliqueSplitter):
                     if proj_i >= self.max_features:
                         break
 
-# XXX: not used right now
+# Test version now
 cdef class MultiViewObliqueSplitter(BestObliqueSplitter):
     def __cinit__(
         self,
@@ -833,11 +833,13 @@ cdef class MultiViewObliqueSplitter(BestObliqueSplitter):
         float64_t feature_combinations,
         const intp_t[:] feature_set_ends,
         intp_t n_feature_sets,
+        const intp_t[:] max_features_per_set,
         bint uniform_sampling,
         *argv
     ):
         self.feature_set_ends = feature_set_ends
         self.uniform_sampling = uniform_sampling
+        self.max_features_per_set = max_features_per_set
 
         # infer the number of feature sets
         self.n_feature_sets = n_feature_sets
@@ -856,6 +858,7 @@ cdef class MultiViewObliqueSplitter(BestObliqueSplitter):
                     self.feature_set_ends,
                     self.n_feature_sets,
                     self.uniform_sampling,
+                    self.max_features_per_set,
                 ), self.__getstate__())
 
     cdef int init(
@@ -867,6 +870,7 @@ cdef class MultiViewObliqueSplitter(BestObliqueSplitter):
     ) except -1:
         Splitter.init(self, X, y, sample_weight, missing_values_in_feature_mask)
 
+
         self.X = X
 
         # create a helper array for allowing efficient Fisher-Yates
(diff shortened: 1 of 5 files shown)


treeple/tree/_oblique_splitter.pyx
```diff
@@ -773,7 +774,7 @@ cdef class MultiViewSplitter(BestObliqueSplitter):
                     feat_i = self.multi_indices_to_sample[idx][ifeature]
 
                     # here, axis-aligned splits are entirely weights of 1
-                    weight = 1  # if (rand_int(0, 2, random_state) == 1) else -1
+                    weight = 1 if (rand_int(0, 2, random_state) == 1) else -1
```
adam2392, line 777: This will change the axis-aligned multi view-splitter I think(?)

adam2392 (commented):
I think this looks correct, but I have not validated it, and also it's usually a bit hard to validate Cython code. 

I would recommend writing a small test Cython test class that you can invoke in Python, which then calls the `sample_proj_mat` for an input array. You can then visualize and verify that it will do what you want it to do for a set of features. You can then manually show that the MultiViewObliqueSplitter samples what you would expect to be the right projection matrix for multiple sets of features.

For example, see the `MultiViewSplitterTester` I wrote for test the MultiViewSplitter, and `plot_multiview_axis_aligned_splitter.py`.

treeple/tree/_oblique_splitter.pyx
YuxinB: Oh, yes, I will change it back, it should be just 1, instead of 1 or -1

YuxinB:
> I think this looks correct, but I have not validated it, and also it's usually a bit hard to validate Cython code.
> 
> I would recommend writing a small test Cython test class that you can invoke in Python, which then calls the `sample_proj_mat` for an input array. You can then visualize and verify that it will do what you want it to do for a set of features. You can then manually show that the MultiViewObliqueSplitter samples what you would expect to be the right projection matrix for multiple sets of features.
> 
> For example, see the `MultiViewSplitterTester` I wrote for test the MultiViewSplitter, and `plot_multiview_axis_aligned_splitter.py`.

Will do!

PSSF23 (commented):
I assume the changes in `forest.py` is to remove the multi-class restrictions? Then we can just remove the assertion line instead of commenting it out.

YuxinB:
@adam2392 Can you help me recheck with Multiview Oblique splitter when uniform sampling condition is true. I modified it to be more reasonable to achieve uniform sampling.

adam2392:
> @adam2392 Can you help me recheck with Multiview Oblique splitter when uniform sampling condition is true. I modified it to be more reasonable to achieve uniform sampling.

Did a loose check, and it looks in the right direction. Will do an in-depth review after we can visually verify the projection matrix makes sense.

adam2392:
I'm unsure why the docs are broken, but it would be nice to be able to check the output visually here: https://output.circle-artifacts.com/output/job/96dcd4fa-4fbc-4b29-b358-314687b9af0b/artifacts/0/dev/use.html via the circleCI job: https://circleci.com/gh/neurodata/treeple/561. It makes reviewing easier/trivial even.

This is what I got though when I ran `plot_multiview_oblique.py` locally. To me this seems weird because your `feature_combinations` is 2, but on average, you're definitely sampling more than 2 feature indices per projection vector. Also, if this is an oblique combination, I'm unsure why the projection weights are only of value +1? For axis-aligned, we make them +1, but for oblique, we made them +/- 1, or 0. Without giving the code a deeper look, I conclude there's something not working. Open to being convinced tho.

Figure_1

YuxinB:
When I directly output the projection matrix from `plot_multiview_oblique.py`, I got matrix:
image
The feature_combination is around 2 on average. And there are -1s, 0s and 1s.

YuxinB:
Done with with visualization bug:
image
image

PR #357, opened 2025-04-22
multitask classifier

CambridgeCat13:


CambridgeCat13:
This PR adds a new Jupyter Notebook to evaluate transfer learning efficiency using the MultitaskForestClassifier (SPORF model). We test transfer from MNIST to FashionMNIST under extremely small target data ratios (0.1%, 0.2%, 0.5%, 1%). And our results show that at 0.2% (16 samples), transfer learning improves accuracy, while at other ratios results are mostly noise.

CambridgeCat13:
New update with MORF and transfer learning using MORF is effective when training data is extremely limited.
The ipynb now is containing both SPORF and MORF.

PR #358, opened 2025-04-22
Allow NaN values in unsupervised random forest

weioren:
Changed _fit_ function in __unsupervised_forest.py_ to add ensure_all_finite = "allow-nan" parameter to ensure NaN values are always allowed.


adam2392:
Hi @weioren note this will not enable splitting on missing values, as the core logic within Cython needs to get implemented.

PR #359, opened 2025-05-13
Fast oblique

ClarkXu0625:
Replace Fisher Yates Shuffle by Floyd's method, a more efficient approach to draw uniform distribution from large ranges, in SPORF (i.e. ObliqueRandomForestClassifier). When the projection matrix is huge (i.e. large number of data features and/or max_features), this update would reduce training time significantly without affecting prediction.

### DIFF
--- a/treeple/tree/_utils.pyx
+++ b/treeple/tree/_utils.pyx
@@ -14,33 +14,6 @@ cnp.import_array()
 from .._lib.sklearn.tree._utils cimport rand_int, rand_uniform
 from libcpp.unordered_set cimport unordered_set
 
-''''
-cdef inline void fisher_yates_shuffle(
-    vector_or_memview indices_to_sample,
-    intp_t grid_size,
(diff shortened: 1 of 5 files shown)


treeple/tree/_utils.pyx
```diff
@@ -41,6 +43,26 @@ cdef inline void fisher_yates_shuffle(
...
+    intp_t k,
+    intp_t n,
+    uint32_t* random_state
+) noexcept nogil:
+    cdef unordered_set[intp_t] seen
+    cdef intp_t i, r, count = 0
+
+    for i in range(n - k, n):
+        r = rand_int(0, i + 1, random_state)
+        if seen.find(r) == seen.end():
```
SUKI-O: Is this right? Shouldn't it be `if seen.find(r) != seen.end():` ?

ClarkXu0625: Oh right, got it fixed

